PR #38, opened 2012-04-06
arrays as values

egraether:
I have seen the discussion about using an array or bezier as parameters (https://github.com/sole/tween.js/issues/8) and then had a look at the spline example (http://sole.github.com/tween.js/examples/05_spline.html)

I thought this feature is something that could be really useful and not too easy to implement on one's own. So it's nice to have it provided from the library.

I accomplished a decent start with linear and spline interpolation for arrays as parameters.
There is an extra check for arrays in the start and update function of each tween, but that shouldn't be an issue at all. It's tested and works with easing. I'm not too sure if API and naming is understandable.

I visualized the different interpolation types in an example.
Next I try to have a look into beziers as well.

What do you think about this?
Is it already too much, or still too little?

thanks @mrdoob for the spline example!

### DIFF
--- a/examples/05_spline.html
+++ b/examples/05_spline.html
@@ -28,34 +28,40 @@
 		<script src="../src/Tween.js"></script>
 		<script>
 
-			init();
+			var target = document.getElementById('target');
+
+			var canvas = document.createElement( 'canvas' );
+			canvas.width = 1024;
+			canvas.height = 512;
+			target.appendChild( canvas );
+
+			var context = canvas.getContext( '2d' );
+			context.fillStyle = "rgb(250,250,250)";
+			context.fillRect( 0, 0, 1024, 512 );
+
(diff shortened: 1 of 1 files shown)


src/Tween.js
```diff
@@ -542,3 +575,87 @@ TWEEN.Easing.Bounce.EaseInOut = function ( k ) {
...
+			var s = 1, i;
+			if ( a[ n ] ) return a[ n ];
+			for ( i = n; i > 1; i-- ) s *= i;
+			return a[ n ] = s;
+
+		}
+
+	} )(),
+
+	CatmullRom: function ( p0, p1, p2, p3, t ) {
```
trusktr: @mrdoob @egraether If using CatmullRom, then we need to supply no more than 4 numbers in an array value? Is there not a more dynamic way?

PR #87, opened 2013-09-06
Add repeatDelay

dnerdy:
repeatDelay is used to control the delay time between each repeat.  This
is useful if, for example, you want to have a 2 second delay between
each repeat but you want the Tween to start immediately. The current
delay method will cause the Tween to delay 2 seconds before starting.

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -100,6 +100,7 @@ TWEEN.Tween = function ( object ) {
 	var _valuesStartRepeat = {};
 	var _duration = 1000;
 	var _repeat = 0;
+	var _repeatDelayTime = undefined;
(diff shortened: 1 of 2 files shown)


dalisoft:
Hi @dnerdy!
What need repeatDelay. delay using as repeatDelay when repeat is used

mikebolt:
Seems to work well. All I had to do was move the tests and update them to the new test format. I also added an extra assert to each test to ensure that the value would stay at the end value during the delay.

We should clearly document the behavior of a tween with both delay and repeatDelay, because it's not intuitive.

dalisoft:
Thanks @mikebolt. Are you get my message on e-mail

PR #88, opened 2013-09-06
Fix Tween synchronization when using repeat.

dnerdy:
The current method of calculating repeat _startTime can cause Tweens to
fall out of sync. The current calculation is dependent on the update
time that triggers the repeat, not the original _startTime and _duration
of the Tween.  This issue can be readily observed by 1) creating two
Tweens with different durations, 2) setting them to repeat, and 3)
starting them at the same time. They slowing fall out of sync.

Fix: _originalStartTime, number of _repeats and _duration are used to
calculate a consistent repeat _startTime. _delayTime is also taken into
account.

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -100,10 +100,12 @@ TWEEN.Tween = function ( object ) {
 	var _valuesStartRepeat = {};
 	var _duration = 1000;
 	var _repeat = 0;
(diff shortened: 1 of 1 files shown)


sole:
There was another fix that might have fixed this issue #187 so I'll close this.

PR #90, opened 2013-09-10
Added the parameter for callback & change the scope

strwind:
1、Added the parameter for callback
2、change the scope of call function to this , instead of _object


mikebolt:
I think that this is out of the question now. It's not a bad idea, but considering that it has been this way for 3+ years, we would break way too much code if we changed this now.

Hopefully users can figure out how to access the original tween via closure, if necessary.

PR #93, opened 2013-10-11
added preservation flag in TWEEN.update to be able to go backward in time

jonobr1:
For the [timeline](https://dl.dropboxusercontent.com/u/1595444/experiments/DAT/timeline/sketches/drawing.html) I'm making with Tween.js I'm serving `TWEEN.update()` my own time variable. This can go forwards and backwards so I need to be able to preserve tweens and not splice them when I'm done with them. My use case is:

```javascript
var t = new TWEEN.Tween(myObject)
  .to({ x: 5, y: 10 }, 750)
  .start(0);

// first tick
TWEEN.update(0, true);

// ...

// second tick
TWEEN.update(1000, true);

// ...

// third tick
TWEEN.update(500, true);
```

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -61,7 +61,7 @@ var TWEEN = TWEEN || ( function () {
 
 		},
 
-		update: function ( time ) {
+		update: function ( time, preserve ) {
 
 			if ( _tweens.length === 0 ) return false;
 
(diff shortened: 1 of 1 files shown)


mikebolt:
Fixed in #271, which is equivalent.

PR #129, opened 2014-07-21
add from method for some use-cases

finscn:
In some use-cases , the object's fields are not  _valueStart ,so maybe user want to do something like this:

```
new TWEEN.Tween( obejct ).from( {x:20, y:20}).to({x:100,y:100}).start();

```


trusktr:
@finscn Unfortunately much time has passed, so this is very out of date with the latest Tween.js file. If you wouldn't mind updating, I think this feature can be convenient. It would be useful, for example, to be able to apply `from` to many objects at once in a `Group`.

trusktr:
@finscn If you decide to update the PR (and add tests), then we can re-open this. Thanks!

PR #147, opened 2014-08-31
Fix onStart to match user guide

BryanMorgan:
The user guide's reference to onStart doesn't work as expected (see http://jsfiddle.net/BryanMorgan/trL3k6j1/ for an example). This change simply resets the _valuesStart object if the onStart callback is fired in case the values are changed inside the callback.

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -286,6 +286,13 @@ TWEEN.Tween = function ( object ) {
 
 				_onStartCallback.call( _object );
 
+				// Reset all starting values, in case they were changed by the callback
+				for ( var field in _object ) {
+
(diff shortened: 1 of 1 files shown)


sole:
Thanks for this PR @BryanMorgan and sorry it's taken me so long to answer it. I missed many pings last year.

I've reviewed your PR and the code, and actually the user guide is wrong. I'm amending it to reflect the actual behaviour of the library--`onStart` is only called *once*, not each time a tween repeats. Its usage is more for syncing other events (e.g. activating things on the document) than resetting values.

sole:
Resolved in https://github.com/tweenjs/tween.js/commit/42387a94665c7ccdf803ddb76a67a6f78302b063

PR #170, opened 2014-11-05
allow FP style callbacks by applying _object

fyyyyy:
this way, its easier for functional programmers to say
tween…..onComplete(Level.destroy).start()

While Level.destroy accepts an object as method, while Level is just a namespace for functions.

Usually in FP, the data is applied as last param for more flexibility, see Ramda.js vs the wrong way in _underscore

https://www.youtube.com/watch?v=m3svKOdZijA

Hope this wont break something else but so far i havent gotten any problems.

Cheers!

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -180,7 +180,7 @@ TWEEN.Tween = function ( object ) {
 
 		if ( _onStopCallback !== null ) {
 
-			_onStopCallback.call( _object );
+			_onStopCallback.call( _object, _object );
 
 		}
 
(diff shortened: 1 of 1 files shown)


mikebolt:
Thanks for the PR. I'm not sure why there was so much hesitation for this simple feature.

I added an explanation to the user guide to document this new feature.

Technically I should not have merged this yet because I did not add tests. This hardly needs testing, but tests will help to ensure that we don't accidentally remove this feature in the future.

PR #182, opened 2015-01-21
fix link to 04_simplest example

JoshuaSP:


sole:
woot! thanks @JoshuaSP and sorry it took so long to merge. I missed many notifications at the beginning of the year.

PR #183, opened 2015-01-27
Tween.speed

JoshuaSP:
Added function for controlling the global speed of all running and future tweens.

Unlike the in-development timescale function, which if applied to a running tween can cause behavior such as running backwards, TWEEN.speed() allows for granular, rapid speed changes across the whole TWEEN environment.

Simply call TWEEN.speed(2) to run tweens twice as fast, TWEEN.speed(0.5) for half as fast, etc.

Included are a test and an example HTML.

Changes made only to Tween.js in src because contributors were asked not to check in tween.min.js. Link in the example HTML is to Tween.js in src - this needs to be changed when tween.js is re-minified. (Should I have done this, too?)

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -63,6 +63,7 @@ var TWEEN = TWEEN || ( function () {
 			var i = 0;
 
 			time = time !== undefined ? time : ( typeof window !== 'undefined' && window.performance !== undefined && window.performance.now !== undefined ? window.performance.now() : Date.now() );
+			time = ( time - this._speedChangeStartAbsolute ) * this._speed + this._speedChangeStartRelative;
 
 			while ( i < _tweens.length ) {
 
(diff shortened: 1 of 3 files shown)


src/Tween.js
```diff
@@ -136,7 +154,18 @@ TWEEN.Tween = function ( object ) {
 
 		_onStartCallbackFired = false;
 
-		_startTime = time !== undefined ? time : ( typeof window !== 'undefined' && window.performance !== undefined && window.performance.now !== undefined ? window.performance.now() : Date.now() );
+		if (time == undefined) {
+
+			time = ( typeof window !== 'undefined' && window.performance !== undefined && window.performance.now !== undefined ? window.performance.now() : Date.now() );
```
dalisoft: * This is drops performance and makes size bit large

src/Tween.js
```diff
@@ -357,12 +386,14 @@ TWEEN.Tween = function ( object ) {
 					_reversed = !_reversed;
 				}
 
-				_startTime = time + _delayTime;
+				_startTime = time;
+				_startTime += _delayTime;
```
dalisoft: * This used little more memory and little size larger

sole:
@JoshuaSP first, apologies for taking a year to get back to this! I missed notifications for a long time and they piled in.

Second, the library has grown to be so complicated I don't want to add new features until I understand what's in there yet. [Full explanation here](http://soledadpenades.com/2015/09/30/tween-js-mega-changes/)

I thank you for your submission but I can't merge it nowadays. Sorry about that! I will keep this in mind for the future.

Issue #145, opened 2014-08-30
Doesn't work in Safari?


matthewcc:
I don't know if this is from a recent update, but tween.js is not working in Safari. The sites in the 'projects using' all seem to be working as expected, but the 'examples' and my own project do not.

PR #185, opened 2015-02-06
Add performance.now polyfill

superlukas:
This should solve the problems people have with tween.js on iOS, as well as improve animation quality in older browsers in general.

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -7,16 +7,28 @@
  * Thank you all, you're awesome!
  */
 
-// Date.now shim for (ahem) Internet Explo(d|r)er
-if ( Date.now === undefined ) {
+// performance.now polyfill
+( function ( root ) {
 
-	Date.now = function () {
+	if ( 'performance' in root === false ) {
+		root.performance = {};
+	}
 
-		return new Date().valueOf();
+	// IE 8
+	Date.now = ( Date.now || function () {
+		return new Date().getTime();
+	} );
 
-	};
+	if ( 'now' in root.performance === false ) {
+		var offset = root.performance.timing && root.performance.timing.navigationStart ? performance.timing.navigationStart
+		                                                                                : Date.now();
 
-}
+		root.performance.now = function () {
+			return Date.now() - offset;
+		};
+	}
+
+} )( this );
 
 var TWEEN = TWEEN || ( function () {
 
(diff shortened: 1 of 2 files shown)


superlukas:
I can fix the failing tests, if you like. I'm not sure tween.js supports non-browser contexts officially anyway.

JosephClay:
+1 Much faster/cleaner than a check on every update

dalisoft:
Before submit please see your code in Chrome Dev-Panel! Your polyfill works normally, but need some  fix!

superlukas:
@dalisoft I did and it worked just fine. Just tell me what's wrong so I can fix it.

sole:
Hi @superlukas - I hadn't seen this PR yet, I'm sorry. Something happened with the notifications.
It looks awesome, I've been meaning to do this for a while so I'm glad you stepped ahead and did that.

I recently moved the project to its own organisation, so I think that might be a reason why Travis isn't too happy. I will look into that. Not sure about errors in Chrome, I haven't tested it...

superlukas:
@sole Yeah I now; don't worry about it :)

The tests fail because `window` isn't available in Node. If you want to support Node as well then just change the naked call to something like  `window ? window.performance.now() : Date.now()`. I didn't choose to not support Node or anything — I just forgot.

I figured there's no point in adding this to the ES5 version anymore given that ES6 is on the agenda now.

sole:
Oh right, I see! That's why they're failing too. Mmm so, I wasn't thinking of canning the es5 version already... there's a lot of people using it and we don't even have an es6 yet :-)

So in order to pass the tests we could either
* add a mini "polyfill" to the tests for them to have a "window" object? although not sure how it'll work with node.js scopes...
* have the tests actually run on a browser e.g. phantom or similar

I think for the sake of running tests locally too, the first option is easier :)

superlukas:
Alright here we go. Don't worry about the mess, I'll squash the commits later. This is not a acceptable solution anyway because of the runtime overhead.

Running the tests in PhantomJS would probably be the best solution although making it work with nodeunit would be way cheaper.

superlukas:
@sole I didn't expect the fix to be that easy. I *think* you can merge the changes now if you like.

sole:
Okay I could finally have a moment to test this one! Looks fine and no errors on Safari / Chrome on Mac. If anyone has a tablet or other iDevice and want to try it out it'd be nice so we can maybe close #145

Merging it in the meantime :) thanks @superlukas !!!

PR #186, opened 2015-02-06
Add better module support (UMD)

superlukas:
I use CommonJS modules, so I don't have a real use case for this, but I figured somebody else might. These are non-breaking changes.

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -754,6 +754,26 @@ TWEEN.Interpolation = {
 
 };
 
-if(typeof module !== 'undefined' && module.exports) {
-	module.exports = TWEEN;
-}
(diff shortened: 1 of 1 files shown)


dalisoft:
I month before added this module defintion, it's native'ly supports defintion. Please re-check your PR!
1) window.TWEEN no need because var TWEEN available.
2) NodeJS unit need for current PR.

superlukas:
1. Yeah I know. I left it in in case Sole wants to go full AMD one day. That's not necessary anymore because she wants to port tween.js to ES6 anyway.
2. Not sure what you mean. Do you want me to add a unit test to check if `require`ing tween.js works? I guess we don't need that anymore either.

PR #187, opened 2015-02-10
Chained tweens get out of sync with sporadic updates

benfoxall:
When a tween has chained tweens, and `TWEEN.update` is called with a value past the duration of tween - the chained tweens are started from that update time, rather than starting at the end of first tween.

```js
// 0 -> 1000
var a = new TWEEN.Tween({})
          .to({}, 1000)

// 1000 -> 2000
var b = new TWEEN.Tween({t:1000})
          .to({t:2000}, 1000)
          .onUpdate(function(){console.log("time: %d", this.t)});

a.chain(b).start(0);

TWEEN.update(1500); // time: 1000 (expected 1500)
TWEEN.update(2000); // time: 1500 (expected 2000)
TWEEN.update(2500); // time: 2000 (expected nothing)
```

I think there is still an issue with repeated tweens, and also when the update gap is longer than the chained tween too; though I'm not 100% sure of the best way of resolving that.


dalisoft:
Hi @benfoxall!
Your want performance slower. Tween.js tweens array removing after completed for performance

benfoxall:
@dalisoft I believe they'll get removed the next time you call TWEEN.update (see [Tween.js#L87](https://github.com/tweenjs/tween.js/blob/master/src/Tween.js#L87))

dalisoft:
This need for performance, TWEEN.update updates active tweens, not a inactive. Tween.js authors best knows what does it!

sole:
Sorry I totally missed this PR. I'm on holidays, but adding this to my agenda for when I'm back.

Thanks @benfoxall and @dalisoft !

sole:
hey @benfoxall I manually ported the fix as it had gone out of sync. Thanks for the PR!

benfoxall:
:+1:

PR #188, opened 2015-02-16
window.performance.now polyfill

dalisoft:
Polyfill for Tween.js


sole:
Hi @dalisoft thanks for the PR but I think we'll go for the other one at https://github.com/tweenjs/tween.js/pull/185 :)

sole:
fixed in #185

PR #197, opened 2015-06-06
Eco-Mode

dalisoft:
Eco-Mode PR - Save your battery up to 50%.
# How it works
* Stops the requestAnimationFrame callback after all tween is completed
* Re-call requestAnimationFrame callback if starting/re-starting tween
* Re-call (reTick) doesn't looping, because toggling by variable ``` _stoppedTick = false|true ```

# Demo Source
```html
<!DOCTYPE html>
<html>
   <head>
      <meta charset="UTF-8" />
      <meta name="viewport" content="user-scalable=no, width=device-width, initial-scale=1"/>
	  <meta name="http-equiv" content="IE=edge;chrome=1"/>
      <title>Demo without stylesheet, #1 test</title>
	  <style>
	  div {
	  width:100px;height:100px;background-color:#0cf;
	  }
	  </style>
   </head>
   <body>
   <div></div>
      <script src="Tween.js"></script>
      <script>
	  function style() {
	  var style = document.querySelector('div');
			for ( var key in this ) {
				style.style[key] = this[key] + 'px';
			}
	  }
        var tween = new TWEEN.Tween({}).to({width:200}, 2000).onUpdate(style).start(),
			tween2 = new TWEEN.Tween({}).to({height:200}, 2000).delay(2000).onUpdate(style).start()
      </script>
   </body>
</html>
```

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -43,7 +43,6 @@ var TWEEN = TWEEN || ( function () {
 			process : function (time) {
 
 				_tick = requestAnimationFrame(TWEEN.process);
-				console.log(time);
 				TWEEN.update(time);
 
 			},
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -32,11 +32,36 @@
 
 var TWEEN = TWEEN || ( function () {
 
-	var _tweens = [];
+	var _tweens = [],
+	_tick,
+	_stoppedTick = false;
 
 	return {
 
 		REVISION: '14',
+
+			process : function (time) {
+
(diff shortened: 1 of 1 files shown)


dalisoft:
Hi @sole.
I maked the modded version of Tween.js and titled Animo.js (https://github.com/dalisoft/animo.js).
Two version have
* First stable v1
* Second now stable v2

But i seen one issue, i think this is a source code issue.
Animation skips every 2nd node when loop

sole:
No, we shouldn't mix `requestAnimationFrame` or even assume that is the way the update function is called.

PR #202, opened 2015-07-12
`sole` -> `tweenjs` for all github references

wblankenship:
Fixes TravisCI badge

Also bower now points to the correct repo and latest version `r14`


sole:
oh thanks, I was almost going to fix this now. I need to make the tweenjs.github.io repo beforehand or none of the examples will work - let me address that before merging this :)

Issue #206, opened 2015-07-17
Removing the 'build' version as it's just such a little improvement vs 'uncompressed'


sole:
This made sense in 2009 or 2010 when the project was started (can't even remember) and people just included `<script>` tags in their code, but nowadays almost everyone uses some sort of build process and frankly if you don't, you probably don't care too much about using a minified or not version.

The last compressed build was about 5k vs about 13k for the original. It's ridiculous to have to spend time on this, and it gets out of sync all so often! (e.g. where this is an issue: #196)

We are now using npm and package.json which allows you to easily install and use this library. The build system essentially just uglifies the code and adds some headers to it so you know where it came from and which revision is it.

I think it would be worth to remove this kind of `dist`ributable version. Any thoughts?

tofumatt:
> nowadays almost everyone uses some sort of build process

I _do_ wonder about this, but if it saves development effort and it's such a small size difference, I'd agree and say "don't offer a compressed version".

I'd love to do that with localForage, but the size diff for it is quite large. I say ditch it!

Issue #155, opened 2014-10-11
ES6


mixonic:
Howdy!

I'd like to use Tween.js in a project, one that will be written in ES6.

Is there any interest in porting Tween.js to ES6? I am happy to undertake this. And happy to answer any questions.

superlukas:
I'm using tween.js with webpack and [6to5](https://6to5.org) without any problems. I'm afraid porting tween.js to ES6 and then transpiling it to ES5 would hurt performance more than anything else.

mixonic:
@superlukas A library being written in ES6 does not mean it must be published as ES6.

Ember.js has been written with ES6 module support for nearly a year now. We still only export the library as a ES3-safe build. Some dependencies, like HTMLBars, are published as ES6, transpiled CJS, and transpiled AMD. This give us great flexibility.

superlukas:
@mixonic I still don't get why you'd need the library to be written in ES6. Native ES6 modules are nice, but tween.js supports CommonJS already which all tranpilers support anyway. I send in a PR yesterday that adds full UMD support.

As I said, I use tween.js with ES6 myself, but I still don't see any reason for the library to be written in ES6. Transpiling, as far as I'm concerned, adds a potential overhead without anybody gaining anything from it. Do you mind telling me what your exact use case is? Do you want to use the new `System` module system or something?

sole:
Hi friends

I just found yesterday that I hadn't got any notification for issues in tween for the last two or three months, and I'm so sorry about that. I don't remember unsubscribing from this, but ugh.

I have been thinking about an ES6 version too--mostly because of the module system. Thanks for bringing this up! :+1: 

I also have been thinking about going *megamicromodular* and split every easing function onto its own module, move tween core to its own module, etc. Publishing the stuff to npm and then assembling the whole thing together in a "full" package. But you could use the functions individually too-maybe you're just interested in them!

But I don't want to be the bottleneck in tween.js so I'm working on moving it to a more community-based project. Therefore I created a project under the tweenjs organisation: https://github.com/tweenjs/es6-tween

I believe the best option here is to be "disruptive" and write this new version from the ground up. With proper version number, transpiler to publish to npm, etc.

I've taken added a number of "desired" features to the issues, please feel free to join in the discussion and let's see if we can shape this together. I guess a good place to start is by answering the first issue: [what's the point?](https://github.com/tweenjs/es6-tween/issues/1)

mixonic:
[...]

PR #214, opened 2015-08-22
Support for tween groups

mikebolt:
In an attempt to fix a bug in another project, I realized that it would be useful if this library provided a way to keep track of separate groups of tweens. I tried out a couple different ways of refactoring the code. The one you see here is fairly minimal. It is based on the observation that the whole TWEEN object is basically just a big global group of tweens. So I just made it into a class called TweenGroup. The only substantial changes are that the Tween constructor can now take a TweenGroup (but defaults to the TWEEN global object, which is now an instance of TweenGroup with the rest of the global properties still thrown onto it), and there is a convenience method on TweenGroup called createTween that does the same thing. The whole purpose of TweenGroups is that they can be updated with separate times, which allows for selectively pausing tweens.

I have added three test cases, and all the tests still pass. I also added example 13 which demonstrates tween groups that can be updated and paused separately.

Please look through the code, and let me know if you have any questions. Thanks!

### DIFF
--- a/examples/13_groups.html
+++ b/examples/13_groups.html
@@ -1,0 +1,175 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<title>Tween.js / stop chained</title>
+		<meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+		<link href="css/style.css" media="screen" rel="stylesheet" type="text/css" />
+	</head>
+	<body>
+		<div id="info">
+			<h1><a href="http://github.com/sole/tween.js">tween.js</a></h1>
+			<h2>13 _ separate tweening groups.</h2>
+			<p>Click a widget to pause or resume.</p>
+		</div>
+		<div style="position: absolute; top: 250px; left: 50px; ">
+			<button id="add">Add Widget</button>
+			<button id="remove">Remove Widget</button>
+         <br/>
+         <br/>
+         <div id="widgetContainer"/>
+		</div>
+
+		<script src="../build/tween.min.js"></script>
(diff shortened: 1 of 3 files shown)


mikebolt:
It looks like Travis CI didn't rebuild tween.min.js, so the new tests I wrote are failing on the minified version. The tests all pass on both the regular and minified versions on my system.

sole:
Hey @mikebolt thanks for your contribution!

It sounds like an interesting observation, I definitely hadn't thought of `TweenGroup`s.

However since you can see our testing situation isn't the best, with the tests that go against tween.min.js failing. There is a bug for removing the extra minified version already: https://github.com/tweenjs/tween.js/issues/206 so I think once that one is sorted out it should be easier to do this kind of changes.

My only worry is that this might make everything a little too complex. I want to make sure the library is still 'simple' and not too many layers of abstractions. Having `TWEEN` be an instance of a group but not really being explicit about it sounds like too obscure. I like the idea of the groups, and it makes sense to me. It also would solve another worry I had regarding TWEEN being this single instance that holds all of the tweens, and being unable to differentiate between them. So that is good, but I'm not sure that architecturally this is the best solution.

sole:
I've been thinking more about this. I think this is something that could be implemented as a stand-alone thing/class if/when the library is split in various files and modules as discussed in https://github.com/tweenjs/es6-tween/issues/1 and https://github.com/tweenjs/tween.js/issues/155

So we could have more than one type of global `TWEEN`. Maybe not everything would need to be a singleton etc.

I'm going to close this by now. Hopefully we can get back to this idea in the future!

PR #215, opened 2015-08-27
Ensure compatibility with ECMAScript 2015 modules

### LINKED ISSUE #209: Performance polyfill does not work when tween.js is used as a module
The performance polyfill is enclosed in a self-executing anonymous function, where ``this`` is passed in as the object to which the polyfill will be attached.

When using tween.js as a module (```TWEEN = require('tween.js');```), ``this`` does not equal ``window`` - so the polyfill won't work.

I assume the current method is used for compatibility with environments that don't have the ``window`` variable - could something like ``typeof window !== 'undefined' ? window : this`` be used instead?

bripkens:
The module evaluation context of ECMAScript 2015 modules is not the
global context. As such `this` does not point to the global `window`
object. This is problematic for users of ECMAScript 2015 modules as the
included `window.performance.now` polyfill will fail to work.

Instead of `this`, the polyfill should be added to the global `window`
object directly. This is unproblematic because the following code
assumes that `performance.now` will be available under the global
`window` object, i.e. the code was never execution environment agnostic.

Furthermore, established polyfills like the one from Paul Irish also
directly register the polyfill on the global `window` object.

Fixes #209

[1] https://gist.github.com/paulirish/5438650

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -8,10 +8,10 @@
  */
 
 // performance.now polyfill
-( function ( root ) {
+( function () {
 
-	if ( 'performance' in root === false ) {
-		root.performance = {};
+	if ( 'performance' in window === false ) {
+		window.performance = {};
 	}
 
 	// IE 8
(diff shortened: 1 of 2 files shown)


CodingFabian:
+1

simonThiele:
+1

sole:
Thank you @bripkens :-)

@CodingFabian and @simonThiele please refrain from +1ing, I have very little spare time for working on tween.js and this just adds noise in my inbox.

Issue #153, opened 2014-10-09
chain method syntax is pretty terrible/unclear


sole:
If you chain one thing it works fine
If you chain two things consecutively only the last one gets chained (?)
If you want to chain several with an array, you need to use `tween.apply(tween, saidArray)` (!?)

that is pretty awkward and error prone!

sole:
Not really, I filed the bug to get the thoughts started :-)

sole:
(also sorry about the delay in coming back to this-I haven't been getting notifications, ugh)

thednp:
The `.chain()` method should create an array of chained tweens and the update function should take care of that.

thednp:
The above solution works fine, but the original code is fine as well.

We can tween multiple tweens like so `tween.chain(tween1,tween2)`. 
So I tested it's fine. We only need to document the chain method.

trusktr:
Nowadays we can easily pass an array with `.chain(...tweensToChain)`, so I'll just close this. Feel free to reply if you have any ideas on how to improve it beyond the builtin syntax.

PR #217, opened 2015-09-18
Fix the chain method

thednp:
As described here. 
https://github.com/tweenjs/tween.js/issues/153#issuecomment-141434478
I also tested this change.


thednp:
The tests need to be updated as well.

PR #223, opened 2015-09-29
(feat) Remove minified version, fixes #206

### LINKED ISSUE #206: Removing the 'build' version as it's just such a little improvement vs 'uncompressed'

sole:
Removed the minified version, also updated all the examples that were using the minified version (!) and tried to erase all references to the minified version in the documentation, etc.


jangdan:
why you removed the minified version? just curious :)

sole:
explained here: http://soledadpenades.com/2015/09/30/tween-js-mega-changes/ :-)

PR #228, opened 2015-10-11
Add tween groups using group keys.

mikebolt:
Hi again. This is another attempt at adding group functionality. It is fully backwards compatible as before. This time there is no group class. Instead, the add, remove, and update functions, as well as the TWEEN.Tween constructor now accept an additional group key as a parameter. If omitted, the group key is naturally undefined, so the default group uses the undefined key. That's all there is to it. It may be a little awkward that the remove function requires the group, but that can be changed easily so that it checks all of the groups instead. It could also be moved into the constructor so that it could reference the group variable through closure. Also note that when emptied, all groups except the default group are deleted. Ensuring that empty groups are deleted prevents memory leaks.

### DIFF
--- a/examples/13_groups.html
+++ b/examples/13_groups.html
@@ -1,0 +1,191 @@
+<!DOCTYPE html>
+<html lang="en">
+   <head>
+      <title>Tween.js / tween groups</title>
+      <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+      <link href="css/style.css" media="screen" rel="stylesheet" type="text/css" />
+   </head>
+   <body>
+      <div id="info">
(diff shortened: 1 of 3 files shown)


dalisoft:
Hi @mikebolt 
Thanks for adding/submitting new PR, but there/this javascript doesn't necessary group. Any developer know how-to use this for tweening as `group`. And just basic users uses other Javascript Animation Engine, not a Tweening Engine (not all, some). or uses CSS3 stuffs.
I improved and utilized Tween.js code and works as Animation Engine.
Thanks.

mikebolt:
Tween groups are now in master.

PR #232, opened 2015-11-02
Add version to package.json

neil-morrison44:
Unable to npm install without the version number in the package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
   "name": "tween.js",
+  "version": "16.3.1",
   "description": "Super simple, fast and easy to use tweening engine which incorporates optimised Robert Penner's equations.",
   "main": "src/Tween.js",
   "homepage": "https://github.com/tweenjs/tween.js",


sole:
The version is added by the semantic-release utility. You can totally npm install tween.js by running `npm install tween.js`:

```bash
$ mkdir tmp
$ cd tmp
$ npm install tween.js
tween.js@16.3.1 node_modules/tween.js
```

neil-morrison44:
https://docs.npmjs.com/files/package.json

Name and version are required fields,
I ran into the error when installing via 'npm install tweenjs/tween.js' as well as when I forked and tried to npm install my repo.

What is the semantic release utility supposed to do?

sole:
The utility runs the tests each time a push is done to the repository, and if all is good, it publishes a new release to npm: https://github.com/semantic-release/semantic-release

The reason why the `version` field is not in `package.json` anymore is so that versions can only be published if the tests pass--this is an automatic process.

I see why you want to install from your fork, but I don't want to bring back the version number to `package.json`, essentially because it's a manual process and I always forget to update / keep it in sync. And since I'm the only maintainer, I want to make sure making a release is as painless as possible for me.

Perhaps you want to clone your fork locally to `node_modules` or any other folder with git submodules; I was going to suggest using npm link but I am not sure if it'll work without the version on the `package.json` file: https://docs.npmjs.com/cli/link - if you try it and it works be sure to let me know.

neil-morrison44:
From the ReadMe on semantic-release it looks like there's a process within that for appending the version number to the package.json:

> 2. semantic-release pre |
> ------------ |
> Based on all commits that happened since the last release, the new version number gets written to the package.json. |

I think having the version number in the package.json is necessary for a few npm features involving updates, but I'm not exactly sure.
I just know that it threw me for a loop for about 25 minutes yesterday so it might be a stumbling block for others too.

sole:
The version number is added to a copy of package.json which is then used to publish to npm. This is automatic.

People who install the package using my provided instructions should not encounter this issue.

sole:
I have been looking into the issues in semantic-release and someone brought this up recently, perhaps we can add a 'bogus' value: https://github.com/semantic-release/semantic-release/issues/116#issuecomment-150829057

neil-morrison44:
Yeah, I think a bogus value would probably save any issues should anyone have any reason to install via github / locally rather than through the npm registered version.

sole:
@neil-morrison44 hey, I've added a version to package.json so you should be able to do installs via git/locally now. thanks for bringing it up!

PR #234, opened 2015-11-16
Don't add tween to root object if root is undefined

Remco47:
In our build environment we force strict mode, leading to an undefined "root". This PR adds an undefined check so we don't get console errors.


sole:
thanks!

PR #235, opened 2015-11-21
Fix bug about "module is not defined"

finscn:
In some environments , there may be a global object named exports. 
for example, some code maybe do this :
```
var exports = exports || window;
```

So ,  I think check module & exports both is a better way.


sole:
Thanks @finscn --sorry about the delay in merging but I hadn't had time to look into this before!

PR #237, opened 2015-12-04
Returning if `window` is undefined.

ibbyya:
In a nodejs environment, window is undefined. The fix makes a return if
`window` is undefined, used in the performance now polyfill.

This is used for isomorphic projects - JS written to run on nodejs and
browser.


mikebolt:
I believe that we now have node.js support in release 16.3.5, so this is no longer necessary.

PR #246, opened 2016-02-12
Allow elastic easing parameters to be set.

salzhrani:


sole:
Thanks!

It would be great to have some sort of demo and documentation too. At the very minimum perhaps add an example using custom elastic easing to the [example 3](https://github.com/tweenjs/tween.js/blob/master/examples/03_graphs.html)?

salzhrani:
Done :)

PR #256, opened 2016-04-08
Update

dalisoft:
* `performance.now` shim removed for IE9 doesn't supports ES6
* `Power Saving` mode implemented
* `Built-In Ticker` implemented, so you can write lesser code and just `tweens` code
* ES6 `let` insteadof `var` for ES6 Browsers

* Sorry for used older version to update

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -1,42 +1,19 @@
 /**
  * Tween.js - Licensed under the MIT license
- * https://github.com/sole/tween.js
+ * https://github.com/tweenjs/tween.js
  * ----------------------------------------------
  *
- * See https://github.com/sole/tween.js/graphs/contributors for the full list of contributors.
+ * See https://github.com/tweenjs/tween.js/graphs/contributors for the full list of contributors.
  * Thank you all, you're awesome!
  */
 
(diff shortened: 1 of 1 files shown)


sole:
I'm sorry but I do not understand. What does this pull request address? What is the purpose of it? I'm sure you'll agree that "Update" is quite ambiguous.

Also we're not using let and ES6 etc yet, and these are so many changes in just one PR. And did you use an old base version?

This doesn't seem very mergeable, I'm afraid.

mikebolt:
Is Tween.js currently compatible with IE9? If not, please open an issue. I can certainly help fix the performance.now polyfill if that is causing the problem.

I am interested in the other improvements, but they should all be committed separately. Please open a separate pull request for each feature, and make your commits on top of the latest version of the code. Also, please document each change clearly in the commit messages and the pull requests.

If you add a new feature, please document it in the user guide.

dalisoft:
So, tween.js keeps code clean and light. I like this. I have own tweening engine (i maked). I tried to add new feature in old PR's, but rejected or not merged. So i don't want to again change something in tween.js

PR #259, opened 2016-04-20
add a-frame as a user of tween.js

ngokevin:


sole:
hey! thanks for the PR and sorry for the delay in addressing it

However, the image does not have the same size as the rest of images, and it will stand out quite weirdly. Do you think you could crop it / resize it to respect the 'style' of the repo?

Thanks

Issue #251, opened 2016-03-20
Easing functions


0x860111:
[From](https://github.com/tweenjs/tween.js/blob/6cb21f23975d0230499a11e567d6c954815dd7f2/src/Tween.js#L598)
```
var a = 0.1;
...
if (!a || a < 1) {
	a = 1;
	s = p / 4;
} else {
	s = p * Math.asin(1 / a) / (2 * Math.PI);
}
```

but _!a || a < 1_ is always **false**.
So function can be just like that, isn't?
```
In: function (k) {
    var a = 0.1;
    var p = 0.4;
    var s = p * Math.asin(1 / a) / (2 * Math.PI);
    return k === 0 ? 0 : k === 1 ? 1 : -(a * Math.pow(2, 10 * (k -= 1)) * Math.sin((k - s) * (2 * Math.PI) / p));
},

```

What is the source of those formulas, by the way?

mikebolt:
I'm assuming you meant that the conditional is always true given that a is set to 0.1 right beforehand, and 0.1 < 1.

This seems to be a problem in the other elastic easing functions as well. The only justification I can think of is that `a` is meant to be a parameter that can be adjusted somehow. Please create a pull request if you can. Keep in mind that the simplified code you wrote is not correct, because the conditional is always true as-is. Also, it may be worth investigating the minified code to see if the minifier is optimizing out that conditional anyway.

The [User Guide](https://github.com/tweenjs/tween.js/blob/master/docs/user_guide.md#available-easing-functions-tweeneasing) says that these are adapted from equations by Robert Penner. I found [this page](http://robertpenner.com/easing/), so they possibly came from one of those libraries or straight from his book.

PR #263, opened 2016-06-04
Issue251

mikebolt:
Simplify the elastic easing functions by assuming values for the parameters. Addresses #251. The elastic easing functions will return the exact same values as before.


sole:
Merged this offline to change the last commit message to a message that would make semantic-release happier :-)

Issue #264, opened 2016-06-10
Removing/stopping+starting tweens inside onComplete callback causes wrong tween deletion


drnlab:
Hi,

First of all, thank you for this awesome lib.

I'm working with tween.js and PIXI in a book-like app with multiple scenes. When changing from one scene to another, this is what happens:
- Start fadeOut...
- When fadeOut is complete:
  - Remove/stop all tweens <--- !?
  - Do some usual stuff like change background, remove/add sprites, stop/play music, etc.
  - Start fadeIn...
- When fadeIn is complete, the new scene is ready.

Not a strange use case, but I'm experiencing what I think is a weird behaviour: after removing or stopping all tweens on fadeOutComplete, fadeIn starts but never updates. I reproduced the issue at: [http://drnlab.net/test/tweenjs/]. Here is the source code:

```
var sprite = { alpha: 0 };

var callbacks = {};

callbacks.onFadeOutUpdate = function() {
    console.log( "onFadeOutUpdate - sprite.alpha = " + sprite.alpha );
};

callbacks.onFadeOutComplete= function() {
    console.log( "onFadeOutComplete.\n\n" );
    //
    TWEEN.removeAll();
    /*
    // Same behaviour stopping all tweens ( please, comment line above - TWEEN.removeAll() - before uncomment this code ):
    var allTweens = TWEEN.getAll();
    for ( var i = allTweens.length - 1; i >= 0 ; i-- ) { // Start from the last index to 0, because tween.stop() changes allTweens.lenght! (?)
        var tween = allTweens[i];
        console.log( "!!! - Stopping 1 tween (fadeOut). Ok!" );
        tween.stop();
    }
    */
    //
    // [...] Do some usual stuff like change background, remove/add sprites, stop/play music, etc.
    //
    console.log( ">>> Start fadeIn!!! - sprite.alpha = " + sprite.alpha );
    tweens.fadeIn.start();
    console.log( "!!! - There is " + TWEEN.getAll().length + " tween to update: our fadeIn. That's ok." );
    console.log( "!!! - And now it suddenly dissapears! :(" );
};

callbacks.onFadeInUpdate = function() {
    console.log( "onFadeInUpdate - sprite.alpha = " + sprite.alpha );
};

callbacks.onFadeInComplete= function() {
    console.log( "onFadeInComplete.\n\n" );
};

var tweens = {};

tweens.fadeOut = new TWEEN.Tween( sprite )
        .to( { alpha: 1 }, 1000 )
[...]

PR #266, opened 2016-06-12
Issue #264

mikebolt:
There's some strange behavior when adding or removing tweens during an onUpdate or onComplete callback.

The problem beforehand was that `TWEEN.update` might remove the wrong tween if a tween occurring before the one being updated was removed.

By using an object to contain all the tweens instead of an array, and by using a unique integer as a tween's key, this problem is fixed.

Tweens that are removed during an update are still updated. Previously this behavior depended on whether the removed tween came before or after the one being updated in the tween array. Tweens that are added during an update will still be updated during the current "frame", just like before. If you add and then immediately remove a tween it should not be updated. In general, don't rely on adding then removing a single tween during the same frame. For more details, read the comment I wrote in `TWEEN.update`.

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -40,7 +40,7 @@ var TWEEN = TWEEN || (function () {
 
 		getAll: function () {
 
-			return Object.keys(_tweens).map(function(tweenId) {
+			return Object.keys(_tweens).map(function (tweenId) {
 				return _tweens[tweenId];
 			});
 
@@ -69,7 +69,7 @@ var TWEEN = TWEEN || (function () {
 		update: function (time) {
 
 			var tweenIds = Object.keys(_tweens);
-		
+
 			if (tweenIds.length === 0) {
 				return false;
 			}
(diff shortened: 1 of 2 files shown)


dalisoft:
works and what fixes. maybe i add in my js

sole:
(For easier referencing)
This is a fix for issue #264

sole:
I'm merging this locally - missing a test or demonstration. I need to log off now for today but if you can't add the test/demo today I will try doing that tomorrow. And publish a new version to npm...

sole:
Actually here's the work-in-progress branch so you can use it instead of it staying on my computer (bad habits die hard): https://github.com/tweenjs/tween.js/tree/hashset-wip

nickli:
@sole @mikebolt 

Oh, to solve this problem does not require the original code to do such a big change can be very simple to solve this problem, I have my project in the repair of this bug. See my pull request [Tween.update's pull request](https://github.com/tweenjs/tween.js/pull/310)

and [issues_309](https://github.com/tweenjs/tween.js/issues/309)

_tweens revised for an Object, and then add a _id, is not the best way, I personally think that before the

 _tweens Array as a more reasonable approach.

nickli:
@sole 

Tween.update is a very high frequency operation, so that compared to the previous practice, Tween.update var tweenIds = Object.keys (_tweens); will bring more than the previous performance overhead.

trusktr:
@mikebolt What about something like this? https://github.com/tweenjs/tween.js/pull/330#issuecomment-288275899

mikebolt:
This issue is now fixed.

PR #271, opened 2016-07-20
added preservation flag in TWEEN.update to be able to go backward in time

sairoutine:
I saw https://github.com/tweenjs/tween.js/pull/93 pull request.
However, the pull request is not merged yet and the owner had gone away.

I need to be able to preserve tweens when It's done.

Thanks.

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -64,7 +64,7 @@ var TWEEN = TWEEN || (function () {
 
 		},
 
-		update: function (time) {
+		update: function (time, preserve) {
 
 			if (_tweens.length === 0) {
 				return false;
@@ -76,7 +76,7 @@ var TWEEN = TWEEN || (function () {
 
 			while (i < _tweens.length) {
 
-				if (_tweens[i].update(time)) {
+				if (_tweens[i].update(time) || preserve) {
 					i++;
 				} else {
 					_tweens.splice(i, 1);
--- a/test/unit/tests.js
+++ b/test/unit/tests.js
@@ -129,6 +129,35 @@
 				test.done();
 
 			},
+			'TWEEN.update() does not remove tweens when they are finished with preserve flag': function(test) {
+
+				TWEEN.removeAll();
+
+				var t1 = new TWEEN.Tween( {} ).to( {}, 1000 ),
+					t2 = new TWEEN.Tween( {} ).to( {}, 2000 );
+
+				test.equal( TWEEN.getAll().length, 0 );
+
+				t1.start( 0 );
+				t2.start( 0 );
+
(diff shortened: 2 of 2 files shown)


dalisoft:
Hi @sairoutine.
I no where seen time goes reverse.
But you can do `reverse` feature for that `tween` you want.
Sorry for bad english, so i think you understand me.

sairoutine:
Hi @dalisoft.
Thank you for your advice.

I don't know `reverse` feature. What is it?

dalisoft:
Hi @sairoutine 
`reverse` feature is tween like this
```javascript
var tween = new TWEEN.Tween({x:0}).to({x:300}, 200).reverse().start();
     tween.onUpdate(function(_object){
     console.log(_object);
// {x:300}
// {x:299}
// {x:298}
// ...
// {x:1}
// {x:0}
     });
```

sairoutine:
hmm.. the code you gave me does not perform correctly.
I can't find `reverse` method in tween's source code.
is there any more approach?

mikebolt:
Looks good, thanks. I just merged this request. I also wrote an extra unit test to make sure that if you update past a finished tween with preserve set to true, you can still call TWEEN.update with a time during the tween and get the expected tweened values. I will commit and merge that extra unit test sometime later.

Could you please update the user guide and documentation to explain how to use this feature, and perhaps add an example? If not, let me know, so I can do it instead.

@dalisoft I think having a chainable reverse feature would be cool. We should add that to the next version. However, this new preserve flag feature is still necessary for what sairoutine is trying to do, I think.

sairoutine:
Thank you for merging!
I was mistaken about dalisoft's proposal, sorry.

>Could you please update the user guide and documentation to explain how to use this feature, and perhaps add an example?

All right! I'm going to make a pull request which update documentation.

Thanks.

dalisoft:
@sairoutine @mikebolt Do you know about GSAP (Greensock Animation Platform) reverse feature? Same feature, i tried add to few year ago, but no merged. Then i maked own JS Animation Engine. Now "Deprecated due of no user". But you can see history and see demo. Thanks

trusktr:
v24.0.0 changed the default value of `Group.update()`s `preserve` parameter to `true`, and is deprecated to be removed in a future major version.

To go in reverse, one must now call `tween.start(startTime)` before applying reverse values. The unit test is updated to show how: 

https://github.com/tweenjs/tween.js/blob/c92f7611052a72bccaf110063c81cec76b6d38b1/src/tests.ts#L152-L155

More details in the release notes:

https://github.com/tweenjs/tween.js/releases/tag/v24.0.0

PR #272, opened 2016-08-01
Replaced unsupported `startsWith` String method on IE that breaks compatibility

sdussaut:
The `startsWith` String method is not implemented in Internet Explorer and though breaks the entire library. Reference here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -316,7 +316,7 @@ TWEEN.Tween = function (object) {
 				// Parses relative end values with start as base (e.g.: +10, -3)
 				if (typeof (end) === 'string') {
 
-					if (end.startsWith('+') || end.startsWith('-')) {
+					if (end.charAt(0) === '+' || end.charAt(0) === '-') {
 						end = start + parseFloat(end, 10);
 					} else {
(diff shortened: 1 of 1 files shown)


mikebolt:
Merged. Thanks sdussaut.

Two questions:

1. How did you discover this issue?

2. Does tween.js pass all tests in IE? (I don't have IE at the moment)

sdussaut:
You're welcome!

I discovered it by simply implementing it on a website for a client :-)
Regarding the tests, I ran them in IE11 and all passed fine except the video one, because only a .webm source is provided (and not supported by IE). To make it work, only a .mp4 source would be needed I guess (but I did not test it at the moment).

Do you want me to add it?

mikebolt:
The examples are good tests, but I was more interested in the nodeunit tests in the test/unit directory. If you open nodeunit.html with a browser, it will display the test results.

No need to add a different video format.

sdussaut:
Sorry for the misunderstanding.
Everything looks fine to me on IE11:
nodeunit-ie11

PR #281, opened 2016-08-10
remove "this" in front of window.

zcherries:
Hello! I noticed this broke your tweening library. window isn't an object of "this" and so if you remove the "this" in front of window, it should fix the library. Thanks for your consideration!

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -72,7 +72,7 @@ var TWEEN = TWEEN || (function () {
 // Include a performance.now polyfill
 (function () {
 	// In node.js, use process.hrtime.
-	if (this.window === undefined && this.process !== undefined) {
+	if (window === undefined && this.process !== undefined) {
 		TWEEN.now = function () {
 			var time = process.hrtime();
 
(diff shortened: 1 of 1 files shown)


mikebolt:
What is the problem you are experiencing? We cannot access "window"
directly without causing a reference error in node.js. The intention of the
code was that "this" would refer to the global object, and it would provide
a safe way of checking for global properties. I suspect that any problems
here are caused by "this" not referring to the global object for some
reason.

If your problem occurs in a browser, please let me know which browser.

mikebolt:
I think your issue was caused by using strict mode. In strict mode, the performance.now polyfill's 'this' value was undefined. [My solution](https://github.com/tweenjs/tween.js/commit/23074dfb946aebca6b583f89ea6f807b7e706c1b) was to bind the polyfill to the global 'this' before calling it.

Thanks for reporting this issue. I probably never would have checked strict mode support otherwise.

zcherries:
That fixes it too. Thanks! Sorry I couldn't respond yesterday.

Issue #267, opened 2016-06-12
Seems like Tween.js won't animate getters/setters.


trusktr:
I don't know why, but in my case the object who's values I'm trying animate isn't being animated. I have code like this:

```js
let positionTween = new TWEEN.Tween(letterNode.position)
    .to(cellPosition, 1000)
    .onUpdate(function() {
        console.log('   final position', cellPositions)
        console.log('     tween value?', this)
        console.log('  actual position', letterNode.position)
    })
    .onComplete(function() { positionTween.completed = true })
    
let task = Motor.addRenderTask(time => {
    if (!positionTween.started) {
        positionTween.start(time)
        positionTween.started = true
    }
    else positionTween.update(time)
    
    if (positionTween.completed) {
        console.log('completed')
        Motor.removeRenderTask(Task)
    }
})
```

The problem is demonstrated by the console.log output.

![screen shot 2016-06-12 at 2 05 23 am](https://cloud.githubusercontent.com/assets/297678/15990232/2b23a3ea-3042-11e6-9ff7-47dca951e8b6.png)

In the screenshot we can see what the final position should be, and what the actual tween values are. For some reason, the actual values aren't changing, and are not approaching the expected end values.

Am I missing something?

trusktr:
I added another console.log statement to show that the object I'm trying to animate is the same object as `this` inside of `onUpdate`:

```js
let positionTween = new TWEEN.Tween(letterNode.position)
    .to(cellPosition, 1000)
    .onUpdate(function() {
        console.log('   final position', cellPositions)
        console.log('     tween value?', this)
        console.log('  actual position', letterNode.position)
        console.log('     same object?', letterNode.position === this) // true!
    })
    .onComplete(function() { positionTween.completed = true })
    
let task = Motor.addRenderTask(time => {
    if (!positionTween.started) {
        positionTween.start(time)
        positionTween.started = true
    }
    else positionTween.update(time)
    
    if (positionTween.completed) {
        console.log('completed')
        Motor.removeRenderTask(Task)
    }
})
```

and the output shows `true`. So, that seems to be all good, but I don't understand why the values aren't changing:

![screen shot 2016-06-12 at 2 18 39 am](https://cloud.githubusercontent.com/assets/297678/15990304/fe53bb1e-3043-11e6-8c5e-99c3de1df2cc.png)

[...]

Issue #244, opened 2016-01-12
Do not interpolate properties that do not exist in the source object


sole:
Given this:

````javascript

var obj = { x: 0 };
var tween = new TWEEN.Tween(obj).to({ x: 100, y: 100 }, 100);
````

Tween.js will try to interpolate the `y` property in `obj`, even if it doesn't exist. Properties that don't exist on the tweened object should be ignored.

PR #291, opened 2016-09-24
Object.defineProperty compatibility

jonobr1:
I love using Tween.js. It's in almost all of my projects like [Patatap](http://patatap.com) to client work like this illustration for [the New Yorker](http://www.newyorker.com/magazine/2015/11/23/doomsday-invention-artificial-intelligence-nick-bostrom). For many of my projects I use a library I created and maintain, [Two.js](http://jonobr1.github.io/two.js). Recently, the two libraries don't seem to be compatible... So, I started to do some digging:

The current version of `Tween.js` isn't compatible with any form of `Object.defineProperty` getter / setters. I modified the `start` method to be compatible with this ES5 paradigm. Consequently, Two.js relies on this paradigm a lot. The line I modified checks `_valuesStart[property] === undefined` which doesn't exist for getter / setters because it's a "function" underneath. However, if you check it on the `_object` instead of `_valuesStart` then it exists. Also, from inferring a few lines down you set `_valuesStart[property] = _object[property]` _anyway_, so I hope  this modification doesn't disrupt other use cases. As per the contribution guidelines I tested all the `/tests` and `/examples` as well as added a test to make sure this wouldn't break in the future.

Hopefully you consider this PR. At the very least, I hope this sparks discussion so that we can make Tween.js compatible with `Object.defineProperty` and libraries like `Two.js` 👍

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -169,7 +169,7 @@ TWEEN.Tween = function (object) {
 
 			// If `to()` specifies a property that doesn't exist in the source object,
 			// we should not set that property in the object
-			if (_valuesStart[property] === undefined) {
+			if (_object[property] === undefined) {
 				continue;
 			}
 
--- a/test/unit/tests.js
+++ b/test/unit/tests.js
@@ -1,7 +1,7 @@
 (function() {
 
 	function getTests(TWEEN) {
-		
+
 		var tests = {
 			'hello': function(test) {
 				test.ok(TWEEN !== null);
@@ -158,25 +158,25 @@
 				test.ok( TWEEN.getAll().indexOf( t2 ) != -1 );
 				test.done();
 			},
-			
-			
+
+
 			'Unremoved tweens which have been updated past their finish time may be reused': function(test) {
 
 				TWEEN.removeAll();
 
 				var target1 = {a:0};
 				var target2 = {b:0};
-				
+
 				var t1 = new TWEEN.Tween( target1 ).to( {a:1}, 1000 ),
 					t2 = new TWEEN.Tween( target2 ).to( {b:1}, 2000 );
 
 				t1.start( 0 );
 				t2.start( 0 );
-				
+
 				TWEEN.update(200, true);
 				TWEEN.update(2500, true);
 				TWEEN.update(500, true);
-				
+
 				test.equal(TWEEN.getAll().length, 2);
 				test.equal(target1.a, 0.5);
 				test.equal(target2.b, 0.25);
@@ -226,7 +226,7 @@
 
 				test.ok( t.onStop() instanceof TWEEN.Tween );
 				test.equal( t.onStop(), t );
-				
+
 				test.ok( t.onUpdate() instanceof TWEEN.Tween );
 				test.equal( t.onUpdate(), t );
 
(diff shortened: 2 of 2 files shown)


mikebolt:
Thanks for this PR. I agree that this is a valuable use case.

I am going to begin reviewing the code and trying to break it :)

mikebolt:
I have added some additional assertions to your new test and it still passes.

However, I noticed that the comment directly above the line you changed described the way it works now, after your change, yet you didn't change that comment. I did a "blame" and discovered that @sole authored the commit that changed that line most recently, without changing the comment. Here's that commit: https://github.com/tweenjs/tween.js/commit/6a6e037e4861683d99f21c87cd394afb2206c17a

It references this issue: https://github.com/tweenjs/tween.js/issues/244

We need to avoid tweening properties that really don't exist, otherwise users may get strange behavior or console errors. Unfortunately we never made a unit test for that issue.

I think we need to use the 'in' operator. MDN says that Object.defineProperty makes non-enumerable properties by default. So our issue is that we aren't detecting non-enumerable properties due to the for-in loop on line 129.

This is quite a puzzle. Remember that a property can be defined and "undefined". Try this code in your console:

    var a = {foo: undefined};
    a["foo"]; // undefined
    a["bar"]; // also undefined
    a["foo"] === a["bar"]; // It looks like the same kind of undefined
    "foo" in a; // true
    "bar" in a; // false

wat

Please read through this related issue, where some of these issues were brought up and some possible solutions were considered: #267

jonobr1:
Thanks for the thorough response. The reason I disregarded that comment is because the line following my check (https://github.com/tweenjs/tween.js/blob/master/src/Tween.js#L178) overwrites whatever the start value is to the object's value _anyway_. So, if the start value was undefined to begin with then no harm no foul... It did pass all the tests ;)

But, I see your point. And further to your point my PR is actually to solve for the inadequacies of the `in` statement on 129. But, I don't know how to solve that nut though off the top of my head. More confusing than the code you wrote is this block.

jonobr1:
So, upon further inspection this is actually on my side of things. There is a property `enumerable` that needs to be set to true in order to be picked up on an `in` statement. e.g:

````javascript
var proto = function() {
  this._foo = 0;
};
Object.defineProperty(proto.prototype, 'update', {
  enumerable: true,
  get: function() {
     return this._foo;
  },
  set: function(v) {
    this._foo = v;
  }
});
console.log('update' in proto); // returns true;
````

I've added that to Two.js and it fixes all my Tween.js woes 👍 Sorry for the false alarm.

mikebolt:
I still think this is an issue in Tween.js. It's not obvious at all. Also, as others have noticed in issue #267, the same issue occurs when defining class methods, because they are not enumerable by default.

Also, I have realized that this solves the problem completely. It does change the behavior slightly - if you remove a property between tween construction and calling `start`, it will no longer try to tween that property... actually that's another bug fix :)

Sorry I got thrown off by Sole's fix for #244. That commit fixed one issue and left a second issue open. This commit fixes that second issue.

This part of the code is messy and undocumented. I would like to refactor it soon.

jonobr1:
Glad it helps!

PR #294, opened 2016-10-12
Fix: TWEEN.now() returns broken timestamp in node.js.

thorikawa:
I found a issue that a tweened value reaches the target value immediately when I use it in node.js. The root cause is that `TWEEN.now()` returns broken timestamp in node.js. Tween.js uses `process.hrtime()` to get the current timestamp but it interprets the returned in the wrong way. Please review and merge my PR.

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -76,8 +76,8 @@ var TWEEN = TWEEN || (function () {
 		TWEEN.now = function () {
 			var time = process.hrtime();
 
-			// Convert [seconds, microseconds] to milliseconds.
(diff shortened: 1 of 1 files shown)


thorikawa:
`process.hrtime()` returns `[seconds, nanoseconds]` according to the [document](https://nodejs.org/api/process.html#process_process_hrtime_time).

mikebolt:
My mistake. It looks like it's been that way since v.0.10. I will pull your
fix sometime today.

milli-, micro-, nano-, pico-, femto-, atto-...

PR #296, opened 2016-11-11
Do not copy every field of the tweenable object, only those which are going to be tweened.

hendrysadrak:
Adds only the values that are set in .to properties, i don't see the point why should every property be added, even if it is not going to be tweened.

It is more of a enhancement than a fix, if there is a reason why it is done like it is right now, could you explain it to me?

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -133,13 +133,13 @@ TWEEN.Tween = function (object) {
 
 		_valuesEnd = properties;
 
-        // Set all starting values present on the target object
-        for (var field in _valuesEnd) {
-        	// only add to updateables if tweenable object has the field
-            if ( typeof _object[ field ] !== "undefined" ) {
-                _valuesStart[ field ] = parseFloat( _object[ field ], 10 );
-            }
-        }
+		// Set all starting values present on the target object
+		for (var field in _valuesEnd) {
+			// Only add to updateables if tweenable object has the field
+			if (typeof _object[field] !== 'undefined') {
+				_valuesStart[field] = parseFloat(_object[field], 10);
+			}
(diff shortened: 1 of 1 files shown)


mikebolt:
Thanks for your interest.

We're already doing that in the `start()` method: https://github.com/tweenjs/tween.js/blob/master/src/Tween.js#L167

It's also slightly different: if it only checked properties on the tweened object, then it would have an issue with non-enumerable properties. We need to wait until we know the target properties in `to()` in order to determine which properties to tween.

However, the code is still confusing. I would like a `_tweenedProperties` array to be calculated in either `to()` or `start()`. The `_valuesStart` object effectively contains a list of the properties to tween, but it could be more clear. Also, if we moved the saving of starting properties into `to()` or `start()` then we could potentially save some memory. Imagine if the tweened object has tons of properties, and the user only tweens two of them. This is probably common. However, it might break existing code. If someone modified the properties of the tweened object between construction and calling `start()`, then it might behave differently after that change.

I will leave this open. If you don't want to keep working on this then you can close it.

mikebolt:
Actually, I think we could just get rid of lines 125-128: https://github.com/tweenjs/tween.js/blob/master/src/Tween.js#L125

That will definitely save memory. Let me test it out.

mikebolt:
This is still happening in two places (which means it was happening in three places before).

https://github.com/tweenjs/tween.js/blob/master/src/Tween.js#L137
https://github.com/tweenjs/tween.js/blob/master/src/Tween.js#L176

I am not sure how I missed this. I think that saving the starting properties should be deferred until `.start()` is called, which is the latest possible opportunity.

PR #297, opened 2016-12-02
Fix lookup of global variables (was broken with Webpack + Babel)

### LINKED ISSUE #292: Tween.js:75 Uncaught TypeError: Cannot read property 'window' of undefined
image

Can't figure out why I'm getting this. I just include Tween.js raw in my bundle. Maybe a guard could bedded checking that `this` is not undefined?

I'm on Chrome Version 53.0.2785.116 (64-bit)

haggholm:
Fixes #292 for me.

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -72,7 +72,7 @@ var TWEEN = TWEEN || (function () {
 // Include a performance.now polyfill
 (function () {
 	// In node.js, use process.hrtime.
-	if (this.window === undefined && this.process !== undefined) {
+	if (typeof (window) === 'undefined' && typeof (process) !== 'undefined') {
 		TWEEN.now = function () {
 			var time = process.hrtime();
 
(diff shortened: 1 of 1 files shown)


webmarkelov:
It works with webpack,  for example: ```javascript import TWEEN from 'tween.js' ```

haggholm:
For me, it threw errors because the `this` that’s being bound is `undefined`. I presume(d) it had to do with Webpack; I don’t really know that for a fact. What I do know is that my build (which uses Webpack and Babel) now works…

mikebolt:
According to the EcmaScript 5 spec, 'this' is bound to the global object:
https://www.ecma-international.org/ecma-262/5.1/#sec-10.4.1.1

Also in the EcmaScript 6 spec:
http://www.ecma-international.org/ecma-262/6.0/#sec-lexical-environments

That means that there's a problem somewhere else.

I wrote the code as it is because I was getting the "ReferenceError: window
is not defined" error in node.js. I was not aware that the "typeof"
operator magically protected against this error. I will test this out over
the weekend, maybe.

mikebolt:
OK, it works with my version of node (v 4.4.4). I also removed the function that wrapped the performance.now polyfill because it is no longer necessary.

PR #304, opened 2016-12-15
Wrong Conversion for Node.js polyfill for performance.now

RobertoDeveloper:
We need to convert seconds straight to nanoseconds in order to add the second item in the array and then divide by 1e6 in order to get milliseconds. (Line: 79)

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -76,7 +76,7 @@ if (typeof (window) === 'undefined' && typeof (process) !== 'undefined') {
 		var time = process.hrtime();
 
 		// Convert [seconds, nanoseconds] to milliseconds.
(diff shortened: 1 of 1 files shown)


mikebolt:
I don't think this is right. It would be correct if you added parentheses:

    return (time[0] * 1e9 + time[1]) / 1000000;

but then it's the same as the current calculation. I got this wrong before but I think it is currently correct.

RobertoDeveloper:
That's right bolt. Also, I saw what the original code is doing. Rather than converting straight to nanoseconds, it's taking the seconds converting it into milliseconds, then converting the nanoseconds into milliseconds, then adding the two values together. 

I like this approach, just read the code wrong the first time.

Thanks for the feedback bolt!!

PR #305, opened 2016-12-20
Updated example: 00_hello_world.html

RobertoDeveloper:
I have updated this 00_hello_word.html in order to include important additions. Such as canceling the animate callback set in requestAnimationFrame, if the the TWEEN.update(time) function returns false.

Also, in the update function, instead of using the position object in the update function. We can use the this key word, since tween.js set's the execution context for this function behind the scenes.


examples/00_hello_world.html
```diff
@@ -47,22 +47,25 @@
...
+			//the this key word refers to the current execution context of the function, in other
+			//words tween.js magically sets this object for you with the correct values.
 			function update() {
-
-				target.style.left = position.x + 'px';
-				target.style.top = position.y + 'px';
-				target.style.webkitTransform = 'rotate(' + Math.floor(position.rotation) + 'deg)';
-				target.style.MozTransform = 'rotate(' + Math.floor(position.rotation) + 'deg)';
-
+				target.style.left = this.x + 'px';
```
trusktr: In the latest version of Tween.js, `this` is refers to the Tween instance, not the animated object. If you can switch back to using `position`, then we can merge this.

PR #310, opened 2016-12-28
 Tween.update in real world, tween.update has chance to remove itself, so we have to handle this situation

nickli:
Tween.update function

in real world, tween.update has chance to remove itself, so we have to handle this situation.
                
in certain cases, onUpdateCallback will remove instances in _tweens, which make _tweens.splice(i, 1) fail


nickli:
In my chart library because of this problem, occasional there will be some very strange phenomenon, the investigation results, because the splice (i, 1) to delete the tween is not the expected results, I temporarily in my js

mikebolt:
Thank you. I think this is a simpler solution.

We need to add some tests.

I don't think line 103 is necessary, but maybe I'm wrong.

nickli:
OK, a few days I prepared a test, I am in my local project to solve this problem 103 lines is the key point, all the problems are when the first line of 98 "var _updateRes = _t.update (time);" After this tween instance object itself does not have a passive remove, it is necessary i ++ or _tweens.splice (i, 1). Maybe I did not say clearly, I still have to prepare a test

nickli:
[https://github.com/nickli/tween.js/blob/master/examples/test_for_issue309.html]test_for_issue309.html)

hi @mikebolt ， I built a simple test html, you can see。

martikaljuve:
Wouldn't it be even simpler to iterate the array in reverse? 

```js
// Inside TWEEN.update
var i = _tweens.length;

while (i--) {
	if (!_tweens[i].update(time) && !preserve) {
		_tweens.splice(i, 1);
	}
}
```

This would change the behavior of TWEEN.update to iterate Tweens in reverse order, unless TWEEN.add used `.unshift`:

```js
add: function (tween) {
	_tweens.unshift(tween);
}
```

mikebolt:
The issue that this PR addresses should now be fixed.

Issue #299, opened 2016-12-10
Group start


romanmarkelov:
I have 30,000 vertices, whether there is a method that will help run the animation for all vertices in a single thread?

my code:
  `_.times` => it is [lodash method](https://lodash.com/docs/4.17.2#times)
```javascript
function createParticle() {
    let tween;
    _.times(30000, n => {
        let vertex = shirt.geometry.vertices[n];
        let particle = particles.vertices[n];
        particle.x = 15 * 20;
        particle.y = 16 * 20;
        particle.z = 14 * 20;

        tween = new TWEEN.Tween(particle)
            .to({
                x: vertex.x * shirt.scale.x,
                y: vertex.y * shirt.scale.y,
                z: vertex.z * shirt.scale.z
            }, 3000)
            .onUpdate(function() {
                particleSystem.geometry.verticesNeedUpdate = true;
            })
            .easing(TWEEN.Easing.Sinusoidal.InOut)
            .start();
    });
}
```

mikebolt:
Could you explain in more detail what you want to do?

mikebolt:
Are you trying to calculate the tweens in a separate thread? We could support this in the future, but right now all tweens are calculated on the main thread.

I think we could accomplish this by allowing the user to initialize the "global" TWEEN object instead of creating it by default in the main thread.

sole:
Or you could run Tween on a web worker. Nothing is stopping you from doing that (as far as I know)

trusktr:
30000 Tweens on 30000 vertices is a bit much. This doesn't slow down for you @webmarkelov? What sort of computer are you running it on?

> Or you could run Tween on a web worker. Nothing is stopping you from doing that (as far as I know)

We'd still need some interface for getting those values back in a way that end users don't actually need to know how to use Web Workers, otherwise the complexity is too high for people knowing enough JavaScript to get up and running with Tween.js. Keeping it in the back of my mind for later...

trusktr:
I do think we need a `Group` feature, where we can control groups of Tweens at once. It would be similar to TWEEN, but not a global, not containing absolutely every Tween every instantiated.

mikebolt:
We have Tween.Group now and WebWorkers are supported.

Issue #260, opened 2016-05-04
support complex tweens


usefulthink:
I did use tween.js a lot in combination with three.js, where there are quite often tweens that need to manipulate properties of different objects at the same time.

What I did mostly was to handle this using a custom onUpdate-function like this:

``` javascript
var object = new THREE.Mesh(/* ... */);
var tweenParams = {rotationX: 0, positionZ: 0};
var tween = new TWEEN.Tween(tweenParams).onUpdate(function() {
  object.rotation.x = this.rotationX;
  object.position.z = this.positionZ;
});
tween.to({rotationX: Math.PI, positionZ: 10}).start();

```

But it would be nice if that could be made a bit easier. 

There are two possible solutions I thought of:

**1) allow tweens with complex objects**

``` javascript
var tween = new TWEEN.Tween(object).to({
  position: { z: 10 },
  rotation: { x: Math.PI }
}).start();
```

**2) something like tweenGroups (i.e. tweens that always start and end together)**

``` javascript
var tween = new TWEEN.Group({
  position: new TWEEN.Tween(object.position),
   rotation: new TWEEN.Tween(object.rotation)
});

tween.to({
  position: { z: 10 },
  rotation: { x: Math.PI }
}).start();
```

The group would expose an interface identical to the regular tweens and forward most calls to the individual tweens.

what do you think?

usefulthink:
I would be very happy to provide an implementation and docs/examples for either variant if you like it.

mikebolt:
I like option 1 better. The only thing it changes is allowing "nested" or "recursive" properties to be tweened. I would also prefer that "tween groups" actually be groups of TWEEN.Tween objects, if/when they are implemented. Synchronous tweens would then be possible with some kind of "to" and "startAll" method of the tween group.

Option 2 is potentially faster than option 1, because option 1 requires checking the recursive properties in the onUpdate method, or at least checking if each property is an object. This will be a very minor performance hit for tweens without nested properties.

Please implement option 1 when you have the time, and make a pull request. Try to change as little of the code as possible and keep the "hot path" efficient.

usefulthink:
@mikebolt I think the first option has a nicer interface, but implementation-wise I would prefer the group-variant, because

 - it's easier to get right (at least i think so)
 - no need to change existing behavior – it could even be implemented externally if desired (thus keeping the complexity and speed of the update-function as it is).
[...]

PR #311, opened 2017-01-01
Tween.group

dalisoft:
@sole, @mikebolt, @webmarkelov, @usefulthink please review this my commit

## Why
* for Fix to: #260, #299.

### NOTE:
* if this will be merged, i would be happy to add new features.
* demo already added

### Examples (and methods comment)
```javascript
let groupTween = new TWEEN.Group({
groupVars1:{x:0},
groupVars2:{y:0}
});

// #to
// tweens {x:0} => {x:50} in 2s and {y:0} => {y:50} in 1s
   groupTween.to({
groupVars1:{x:50},
groupVars2:{y:50}
}, {
groupVars1: 2000,
groupVars2: 1000
}); // same as TWEEN.Tween.to

let tweenObj = {j:0};
let tween1 = new TWEEN.Tween(tweenObj).to({j: 100}, 1000);

// #insert
groupTween.insert(tween1); // adds to grouping and tweens together with TWEEN.Group tweens

// #easing, #interpolation, #delay, #repeat => #METHOD_STATIC
groupTween[METHOD_STATIC]({
groupVars1: TWEEN.Easing.Elastic.InOut,
groupVars2: TWEEN.Easing.Bounce.Out
})

//#start
groupTween.start(time) // time: all defined in-group tweens, does same as TWEEN.Tween#start(time) but for all tweens

//#stop
groupTween.stop() // stop all tweens defined in group

//#filter
groupTweens.filter(tween => tween.isPlaying()); // removes paused/stopped tweens from in-group tweens list

//#erase
groupTween.erase(tween1) // finds tween1 from group tweens then removes from it

//#onUpdate
groupTween.onUpdate(function(object, elapsed){
// object => total all mixed properties tween object
// elapsed => total elapsed number ( 0 ... 1 ) with calculating all tweens elapsed it gives overall elapsed of group tween
})
```

### DIFF
--- a/examples/13_group.html
+++ b/examples/13_group.html
@@ -29,8 +29,8 @@
 	<div id="container">
 	
 	</div>
-	 <script src="../test-case/Tween.s.js"></script>
-      <script src="Tween.Group.js"></script>
+	 <script src="../src/Tween.js"></script>
+      	  <script src="../src/Tween.Group.js"></script>
       <script>
 
 			var id = document.getElementById('container');
--- a/examples/13_group.html
+++ b/examples/13_group.html
@@ -1,0 +1,72 @@
+<!DOCTYPE html>
+<html>
+   <head>
+      <meta charset="UTF-8" />
+      <meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=no">
+      <title>Tween.Group.js - Test #1</title>
+	  <style>
+	  html, body {
+	  font-family: Segoe UI, Arial;
+	  }
+
+	  #container {
+		width: 400px;
+		height: 300px;
+		margin: 5px auto;
+		padding: 20px;
+		background: #ccc;
+	  }
+	  
+	  code {
+		background: #eee;
+		padding: 10px;
+	  }
+	  </style>
+   </head>
+   <body>
+
+	<h1>TWEEN.Group Demo</h1>
+	<div id="container">
+	
+	</div>
+	 <script src="../test-case/Tween.s.js"></script>
+      <script src="Tween.Group.js"></script>
+      <script>
+
+			var id = document.getElementById('container');
+			var j = {
+				j: 0
+			};
+			var t1 = new TWEEN.Tween(j).to({
+					j: 100
+				}, 1000);
+			var tween = new TWEEN.Group({
+					a: {
+						x: 0
+					},
+					b: {
+						y: 0
+					}
+				}).to({
+					a: {
+						x: 50
+					},
+					b: {
+						y: 30
+					}
+				}, {
+					a: 500,
+					b: 800
+				}).insert(t1).onUpdate(function (val) {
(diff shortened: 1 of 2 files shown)


usefulthink:
First of all, I am not a maintainer of the project, so this is kind of an outside view. It is also not an exhaustive review: I didn't really look into the use-cases covered by your implementation and the API-surface it provides. This is more about the coding-style and overall structure.

In general, you should really try to adjust yourself to the coding-style used in the project. Two points to make here:

 - tween.js has it's UMD-Wrapper at the bottom and uses a completely different structure for the source-file (for instance how re-initialization is prevented). While this might be a matter of taste (there is no one solution that is subjectively "better" than another) it is the taste and preference of the creators and maintainers that counts. You might disagree, but it's their project so there's that.

 - you are writing in your commit-summary that you are not interested in node. However this project is written to be compatible with node, so you really can't just ignore that.

Then some things about the code:

 - there are several helper-functions defined (`var _helper = {...}`), some of which aren't even used and the others being used just a single time. I think it might be better and easier to read to inline these functions. 
 - basically the same goes for the `_ev.emit`-function. It is only used to be assigned to `TWEEN.Event.Emitter` (btw. if you're using the naming from node it'd be better to also adapt the semantics, otherwise people will be confused) but that in turn is never used. Not sure why you added it, but it apparently has nothing to do with the `TWEEN.Group`-implementation.
 - I would recommend against using abbreviated names for variables (use `object` instead of `obj`, `value` instead of `val`). It generally improves readability and removes ambiguity.
 - some variable- and method-names should be improved. For instance, the function-name `listenTweens` suggests it has something to do with events, where all it does is `_listOfTweens.forEach()` (as said before, helper-functions like this one should probably better be inlined). Think about what someone might think the variables contain without having read all the code. Another example: You are using names like `getDuration` or `getValue` for simple values. Those names suggest to me they are functions (I think I have to call `getValue()` to get the actual value).
 - you are mixing variables with underscore-prefix and without with no apparent reason.
 - generally, I find the code a bit confusing and after reading it for like 20 minutes i still have no clear idea what it does, why it does that and how its done (this wasn't the case for tween.js itself).

dalisoft:
Like before my PRs this is too not reviewed nor merged. So i close this. Get, change, adapt and improve myself is better, faster and effecient than wait a PR

mikebolt:
I don't see anything in this PR that will break node.js. We might have to change a few things to support having separate source files, but that's not an issue.

We will need to add some tests and some documentation. I can help with that. Could you briefly explain what each method does?

I think the most important feature is the ability to control tween groups separately. That was the original issue I had with tween.js that brought me to look at the source code and start helping out with this project. I would like to see an example where you can control two different "groups" of tweens separately, for example with a start and a stop button for each group. Instead of using TWEEN.update, we will have to use an "update" method of the tween group. I think a simple for-each will work, where inside the loop you use each tween's update method.

dalisoft:
@mikebolt see updated comment (first commend), which includes small comment about what it does, but about tests i still learning.

mikebolt:
Tween groups are now in master.

flavia-it:
@mikebolt This Ticket has nothing to do with #346. @dalisoft and @alexburner just used the same name for their ideas...

alexburner:
Oo yeah I had been wondering about that: https://github.com/tweenjs/tween.js/pull/346#issuecomment-311551071

Issue #78, opened 2013-06-27
Tweening arrays and objects?


gautamadude:
So I have these "from" and "to" objects that I need to tween:

```javascript
var from = {x: 0, y: {a: 0, b: 0}, z:[0, 0], color: "#000000"};
var to = {x: 10, y: {a: 20, b: 30}, z:[40, 50], color: "#ffffff"};
```

Tween.js doesn't handle the "y", "z" or "color" values out of the box, but would it make sense to do so? Maybe not colors, but what about objects and arrays? And perhaps even nested objects and arrays like so:

```javascript
{x: 0, y: {a: {t: 0, k:0}, b: 0}, z:[[5,2], [10]], color: "#000000"};
```

So far to deal with this I came up with the following in a hurry, and I still need to implement support for nested arrays and objects (also if you know of a better way I'd be interested):

http://jsfiddle.net/gautamadude/YRbQ3/
```javascript
var from = {x: 0, y: {a: 0, b: 0}, z:[0, 0], color: "#000000"};
var to = {x: 10, y: {a: 20, b: 30}, z:[40, 50], color: "#ffffff"};
var speed = 2000;
var easing = TWEEN.Easing.Linear.None;
var delay = 0;
tweenWithObjectArrayAndColorSupport(from, to, speed, easing, delay);

function tweenWithObjectArrayAndColorSupport(from, to, speed, easing, delay) {
	var outFromTo = tweenSeperate(from, to); //Seperate the colors, array objects, and objects
											 //from the "from" and "to" objects.

	//Tween the remainder
	new TWEEN.Tween( from )
		.to( to, speed )
		.easing( easing )
		.delay(delay)
		.start();

	//Create Tweens for the separated values
	for (type in outFromTo.from) {
		for (attr in outFromTo.from[type]) {
			var vals = {"attr": attr, "type": type};
			createTween(vals, outFromTo.from[type][attr], outFromTo.to[type][attr], speed, easing, delay);
		}
	}
}

function tweenSeperate(from, to) {
	var vals = {};
	var objs = [from, to];
	for (o in objs) {
		var outColors = {};
		var outObjects = {};
		var outArrays = {};
		for (a in objs[o]) {
			if (typeof objs[o][a] == "string") {
				if (/(^#[0-9A-F]{6}$)|(^#[0-9A-F]{3}$)/i.test(objs[o][a])) {
					var t = tinycolor(objs[o][a])
					outColors[a] = t.toHsv();
					delete objs[o][a];
[...]

Issue #150, opened 2014-09-17
Using relative values in 'array of values' tween


sinujohn:
Currently it is not possible to use relative values (string values like "+50") when tweening to arrays of values.
For example, if relative values were allowed, the following code makes a sprite jump 3 times:

``` javascript
var tween3 = new TWEEN.Tween(position)
                .to({x: ["+50", "+50"],
                    y: ["-50", "+50"]})
                .repeat(2)
                .onUpdate(function() {
                    bunny.position.x = position.x;
                    bunny.position.y = position.y;
                });
tween3.start();
```

Can this be added in a future release?

yqrashawn:
Still not added?

sole:
@yqrashawn no. Do you want to chime in and contribute some code? That would be lovely 😃

yqrashawn:
Maybe next month, when fininsh my project. Maybe still busy next month.

mikebolt:
@sole Do you think it would work to convert arrays with relative values to
non-relative arrays first? If so, this wouldn't be difficult. Probably 5
lines of code all in one place.

sole:
@mikebolt but the relative values are computed each time the tween is re-started, how would you account for that? Sounds a bit more than 5 lines of code but maybe I'm overengineering!

trusktr:
This was recently added and tested!

trusktr:
> but the relative values are computed each time the tween is re-started, how would you account for that? Sounds a bit more than 5 lines of code but maybe I'm overengineering!


This is currently the case, but when I was looking at the code I saw a way we can just apply values each time the tween starts. Coming soon...

PR #313, opened 2017-01-03
Add tween to nested object

FlorianBELLAZOUZ:
Hi there !

In reaction to this issues  #52 #78, I sugest this simple API :

With this PR Tween.js can now change properties across nested object, to do this you just need to use a string path pointing to the properties. For example:

```javascript
var nestedObject = { scale: { x:0, y:0 } }

var tween = new TWEEN.Tween(nestedObject).to({ 'scale.x': 100, 'scale.y': 100 });
```

This PR includes UnitTests and Documentation.

I'm open to any sugestion.

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -364,7 +364,7 @@ The way these values are calculated is as follows:
 * the progress (from 0 to 1) is used as input for the interpolation function
 * based on the progress and the array of values, an interpolated value is generated
 
-For example, when the tween has just started (progress is 0), the interpolation function will return the first value in the array. When the tween is halfway, the interpolation function will return a value approximately in the middle of the array, and when the tween is at the end, the interpolation function will return the last value. 
+For example, when the tween has just started (progress is 0), the interpolation function will return the first value in the array. When the tween is halfway, the interpolation function will return a value approximately in the middle of the array, and when the tween is at the end, the interpolation function will return the last value.
 
 You can change the interpolation function with the `interpolation` method. For example:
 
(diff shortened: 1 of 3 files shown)


FlorianBELLAZOUZ:
I add the posibility to use array tween with nested object.

```js
var obj = { scale: { x: 0 } };

var t = new TWEEN.Tween( obj ).to( { 'scale.x': [ 50, 100 ] }, 100 );
```

By the way i fix and unitTest the issues #150

```js
var obj = { scale: { x: 100 } };

var t = new TWEEN.Tween( obj ).to( { 'scale.x': [ '+50', '+100' ] }, 100 );
```

dalisoft:
It looks nice. I test it. If works i add this to my library and give you credit author note

dalisoft:
Very nice. It works, but calling on every frame is just performance-killer

FlorianBELLAZOUZ:
Hi @dalisoft thanks,

Can you provide some code and benchmark ?

mikebolt:
Thanks for the PR. I think this is a much-needed feature. It can help with tweening certain Three.js objects. I like the API with dot notation. We might also want to support using a nested object as the target without the dot notation, for example:

    var t = new TWEEN.Tween({rotation: 5, position: {x: 4, y: 8}}).to({rotation: 12, position: {x:5, y:5}}, 4000);

I am worried about performance in the standard use-case, where tweens are not grouped. I think we will need to perform some tests to make sure. If there is a performance hit, then I have a few ideas for making it more efficient.

I am not worried about performance when using nested tweens. If it's a bit slower, that's OK. But we don't want to make existing code slower.

trusktr:
@FlorianBELLAZOUZ Thanks for bringing up the good idea for nested tweening, however closing in favor of https://github.com/tweenjs/tween.js/pull/366.

PR #319, opened 2017-02-13
Implement onBeforeRepeat callback

jshawl:


trusktr:
This is a good event hook idea.

We should implement a basic event pattern to let any number of listeners hook into the events, unsubscribe, etc. Until then, I think it can be useful to have this.

trusktr:
@jshawl If you can update and fix conflicts, and also test for counter to be incremented to `2` (not just up to `1`) then let's get it merged.

mikebolt:
This was added in another PR. We now have onRepeat callbacks. See the user guide for reference.

PR #325, opened 2017-03-19
perf: improve memory usage by storing methods on the Tween prototype

trusktr:
Also keeps private variables private through a WeakMap.

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -98,236 +98,250 @@ else {
 	};
 }
 
+var privates = new WeakMap();
+
+function _(instance) {
+	var privateObject = privates.get(instance);
+
+	if (!privateObject) {
+		privateObject = {};
+		privates.set(instance, privateObject);
+	}
+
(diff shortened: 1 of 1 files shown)


trusktr:
First time using semantic-release, so hoping I did the commit message right.

trusktr:
This reduces memory usage (for storing class methods) from O(n) to O(1), where n is the number of Tween instances.

**Before:** 100 `Tween` instances -> 1700 `Function` instances
**After:** 100 `Tween` instances -> 17 `Function` instances

(there are currently 17 methods on the Tween class)

trusktr:
@mikebolt @sole @dalisoft

Closing this one in favor of https://github.com/tweenjs/tween.js/pull/325 which is compatible with older browsers, we can add new language features after https://github.com/tweenjs/tween.js/pull/327.

trusktr:
cc @sole @mikebolt @dalisoft

PR #326, opened 2017-03-19
perf: improve memory usage by storing methods on the Tween prototype

trusktr:


trusktr:
This reduces memory usage (for storing class methods) from O(n) to O(1), where n is the number of Tween instances.

**Before:** 100 `Tween` instances -> 1700 `Function` instances
**After:** 100 `Tween` instances -> 17 `Function` instances

(there are currently 17 methods on the Tween class)

dalisoft:
With es6-tween this things does automacilly if i am not wrong

mikebolt:
I am definitely in favor of this. I will check it out soon.

Do you not have maintainer access yet, or are you just waiting for a code review?

trusktr:
@mikebolt No maintainer access yet, so just waiting for review, but even with maintainer access I think it's a good idea to make PRs so multiple maintainers can discuss and approve.

src/Tween.js
```diff
@@ -99,235 +99,249 @@ else {
...
+	this._chainedTweens = [];
+	this._onStartCallback = null;
+	this._onStartCallbackFired = false;
+	this._onUpdateCallback = null;
+	this._onCompleteCallback = null;
+	this._onStopCallback = null;
+
+};
+
+TWEEN.Tween.prototype = assign(Object.create(Object.prototype), {
```
martikaljuve: I think this could be simplified to just:

`assign(TWEEN.Tween.prototype, {`

(similar to how THREE.js does it: https://github.com/mrdoob/three.js/blob/dev/src/core/Raycaster.js#L62)

mikebolt:
My benchmarks suggest that this PR alone leads to a factor-of-5 speedup and memory reduction. These benchmarks involved creating 1 million tweens and calling TWEEN.update(). I will upload these benchmarks later.

I see a minor improvement in FPS on the bars_01 example in Chrome.

trusktr:
Cool!

I will be more available starting in July. :}

src/Tween.js
```diff
@@ -99,235 +99,249 @@ else {
...
+function assign(target, source) {
+	var keys = Object.keys(source);
+	var length = keys.length;
+
+	for (var i = 0; i < length; i += 1) {
+		target[keys[i]] = source[keys[i]];
+	}
+
+	return target;
+}
```
trusktr: @mikebolt Hmmm, I forgot why I did this in place of Object.assign. I don't usually do things just for the heck of it though. Next time I'll be sure to note reasons for things.

trusktr: Oh right, because older browsers don't have Object.assign. A transpile step can help here. There's also http://polyfill.io, which is really nice.

PR #330, opened 2017-03-21
Update tweens in reverse. Fixes #264.

### LINKED ISSUE #264: Removing/stopping+starting tweens inside onComplete callback causes wrong tween deletion

martikaljuve:
This is a possible fix for handling the case where a Tween is removed during its update. 

By having TWEEN.update iterate the Tweens in reverse, a hash map is not required, unlike #266. This has the side effect that Tweens started during the current TWEEN.update (including chained tweens) won't get updated until the next TWEEN.update. This could probably be added by updating an array of `addedTweens` in TWEEN.add and processing them in TWEEN.update, but I personally don't see the need for it.

Any thoughts?

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -47,16 +47,18 @@ var TWEEN = TWEEN || (function () {
 				return false;
 			}
 
-			var i = 0;
+			var i = _tweens.length;
 
 			time = time !== undefined ? time : TWEEN.now();
 
-			while (i < _tweens.length) {
+			while (i--) {
 
-				if (_tweens[i].update(time) || preserve) {
-					i++;
-				} else {
-					_tweens.splice(i, 1);
+				var tween = _tweens[i];
+
+				if (!tween.update(time) && !preserve) {
+					if (_tweens.indexOf(tween) !== -1) {
+						_tweens.splice(i, 1);
+					}
 				}
 
 			}
@@ -67,7 +69,6 @@ var TWEEN = TWEEN || (function () {
 	};
 
 })();
-
 
 // Include a performance.now polyfill.
 // In node.js, use process.hrtime.
--- a/test/unit/tests.js
+++ b/test/unit/tests.js
@@ -938,7 +938,7 @@
 
 			},
 
-			'Test TWEEN.Tween.chain progressess into chained tweens': function(test) {
+			'Test TWEEN.Tween.chain progresses into chained tweens': function(test) {
 
 				var obj = { t: 1000 };
 
@@ -949,6 +949,9 @@
 				var next  = new TWEEN.Tween(obj).to({ t: 2000 }, 1000);
 
 				blank.chain(next).start(0);
+
+				TWEEN.update(1000);
+				test.equal(obj.t, 1000);
 
 				TWEEN.update(1500);
 				test.equal(obj.t, 1500);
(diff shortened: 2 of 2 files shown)


trusktr:
Iterestingly, this would be solved if using `Set`, where you can remove elements from a Set while iterating and it will still iterate over all of the remaining ones with no skipping.

Using Set would also prevent the skipping-to-next-update of started and chained tweens.

Unfortunately, if we support ES5, we have to use Babel to use Set.

I think I'm in favor of #266 if that one behaves as as expected in these edge cases.

trusktr:
What about just

```js
var currentTweens = _tweens

_tweens.length = 0

var tween
while (tween = currentTweens.shift()) {

    tween.update();

}
```

If a tween adds another tween, it will start in the next frame.

Does that solve the problem?

martikaljuve:
I also thought about using _tweens.slice() similar to your last suggestion, but wouldn't that create a lot of arrays for the garbage collector?

mikebolt:
The HashSet code is in master now so this is no longer necessary. Thanks.

PR #331, opened 2017-03-21
Split tween.js into separate files and bundle it with rollup.

martikaljuve:
Similar to #327, but using [Rollup](http://rollupjs.org/) + [Bublé](https://buble.surge.sh/guide/) instead of babel to get a minimal bundle.

* changed jscs and jshint to run on the whole `src` dir.


trusktr:
* I think this PR is too big. Let's first settle on a transpilation step (Babel vs Buble, never saw Buble before, am going to check it out). Can you make a new PR with just Buble, no file separation? I think it would be easier that way so we can compare Babel vs Buble, then move on to Rollup vs Webpack vs Browserify vs etc.
* I like the concept of rollup, but if you use it in bigger projects you'll find that it may still [have bugs](https://github.com/rollup/rollup/issues/1018). Babel's (and Webpack's) ES6 module is more solid.

trusktr:
I like the file separation though! I just think it may be better to introduce smaller changes. I'm reading up on Buble now, looks cool.

trusktr:
The output of Buble looks.... _really nice_.

The question is, is it as mature as Babel? For example, try this in Buble repl:

```js
let s = foo`
  hello ${name}!
  it's good to see you`;
```

trusktr:
After thinking about it for a moment, I'm in favor of merging this commit because the output is cleaner than Babel (with acceptable limitations), the file separation is really nice, and we most likely won't run into any rollup bugs in this case.

Would the commit message need to be modified so that it follow semantic-release format? f.e. `refactor: split tween.js into separate files and bundle it with rollup`

martikaljuve:
For reference, here's the bundled dist/TWEEN.js file that Rollup creates: https://gist.github.com/martikaljuve/a0629891480245aa31fc57f6724bbf7d

trusktr:
@martikaljuve It is clean output!

trusktr:
I'm trying to switch a [project of mine](https://github.com/trusktr/infamous) to Buble, but no luck: https://gitlab.com/Rich-Harris/buble/issues/206

mikebolt:
I have used gulp before so I might try gulp.

trusktr:
@martikaljuve I finished [switching infamous to Rollup+Buble](https://github.com/trusktr/infamous/releases/tag/v17.0.5), the result is awesome (the lean and clean output code).

Can you resolve conflicts so we can move forward with the idea here?

mikebolt:
I just merged in another PR for Rollup.

PR #332, opened 2017-03-22
docs: use a table for examples in readme.

martikaljuve:
See https://github.com/martikaljuve/tween.js/tree/readme-examples-table#examples for example. I think this looks better than the current [inline](https://github.com/tweenjs/tween.js/tree/ae085330fbc0c4e7911c69e786bd3f697f95943f#examples) solution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,20 +98,139 @@ Then reference the library source:
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


sole:
It does! Thank you, it is great =)

PR #349, opened 2017-06-30

sole:
It was pointing to the old package (tween.js instead of @tween.js/tween.js)


sole:
argh closed, it is using the wrong base

PR #350, opened 2017-06-30

sole:
it was pointing to the old package


sole:
don't merge this-trying to investigate what's up with travis

PR #351, opened 2017-06-30

sole:
it was pointing to the old package (hopefully this should work with semantic-release)


sole:
Going to merge this and hope it allows semantic-release to work by giving it its tokens =)

Issue #353, opened 2017-07-04
[doc] clarify repeat


andreasplesch:
As far as I can tell the repeat argument is the number of _additional_ tweens after the first tween. So .repeat(10) actually means running the tweet 11 times.

Since it is not possible anymore to change the semantics of repeat to match the probably more commonly expected meaning, the documentation could explain the behavior explicitly.

PR #354, opened 2017-07-04
[doc] clarify repeat behaviour

andreasplesch:
addresses #353 by explaining that the repeat argument is the number of tweens after the initial tween.


mikebolt:
Thanks!

PR #356, opened 2017-07-14
Expose isPlaying

siwel:
It can be useful to know if the tween is started, while the user could keep track of this themselves we already track this so I have added a getter to expose this.

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -153,6 +153,10 @@ TWEEN.Tween.prototype = assign(Object.create(Object.prototype), {
 	getId: function getId() {
 		return this._id;
 	},
+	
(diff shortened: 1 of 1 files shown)


mikebolt:
I think there's an issue with _isPlaying. It will be true after a tween finishes. Once that issue is fixed I can merge this.

mikebolt:
I made a small change to fix that issue, but there is still another issue. isPlaying() will return true after .start() is called, but before the tween actually starts, if a later start time is specified. It is not straightforward to fix this because the stop() method relies on _isPlaying and the stop() method is supposed to work even before the tween actually starts.

PR #363, opened 2017-07-27
feat(startTime)

### LINKED ISSUE #345: Allow relative values for tween.start()
`tween.start('+1000')` would start a tween 1 second from now, and `tween.start('-1000')` would start a tween 1 second ago, which would effectively skip the first second of the tween.

This is different from `delay(1000)` because it only applies the delay one time.

dalisoft:
now `start` method time argument can accept value relative to current `time`
closes #345

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -172,7 +172,7 @@ TWEEN.Tween.prototype = {
 
 		this._onStartCallbackFired = false;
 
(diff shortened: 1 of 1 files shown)


mikebolt:
I added example#13, because this is difficult to test with nodeunit.

PR #364, opened 2017-07-27
fix(onUpdate): fix for ES6

### LINKED ISSUE #359: provide context to callback
Is it possible to provide context to the callback for onUpdate, onComplete? For example,

```
.onUpdate( function(ctx){

})
```
Much clearer then using `this.` and also doesn't conflict with es6 fatarrow.

dalisoft:
just not breaking change, because a lot of peoples uses this app and they can't make change for "10-15MB" of code
closes #359

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -380,7 +380,7 @@ TWEEN.Tween.prototype = {
 		}
 
 		if (this._onUpdateCallback !== null) {
-			this._onUpdateCallback.call(this._object, value);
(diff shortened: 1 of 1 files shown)


mikebolt:
I am planning to work on this today. I will make a PR for version 17 and ask for your feedback.

dalisoft:
Thanks

mikebolt:
I went ahead and made the breaking change for v17. If anyone reading this wants the 'value' parameter back, please let us know.

Issue #367, opened 2017-09-14
Stopping a tween during onUpdate throws an exception


iangilman:
If I have two tweens running, and during the onUpdate for one, I call `stop` on the other, this line throws an exception:

https://github.com/tweenjs/tween.js/blob/6da570fe81857416dec1f746a20466b185849a99/src/Tween.js#L64

Seems like we should just check whether `this._tweens[tweenIds[i]]` is undefined before calling `update` on it. 

Also, the next line throws an exception if you call `stop` on a tween during its own `onUpdate`, so it could use a guard as well.

I'm happy to make this patch if it seems appropriate.

mikebolt:
Ideally you can avoid calling stop within onUpdate, and instead determine
whether to stop the tweens within requestAnimationFrame or your main update
function.

The reason I recommend this is because the update order of the tweens is
undefined. If you stop tween A within the update of tween B, then tween A
may or may not be updated for the current frame, depending on the order.

I think you have a good solution and I would appreciate a PR.

iangilman:
Yes, that makes sense. I think it's good to have the guards in place though, since it's easy to trigger. I've made #368 to fix.

mikebolt:
Wow, GitHub automatically closed this issue...

Fixed in 17.1.1

iangilman:
Yeah, I did that by putting "fixes #367" in the pull request description (I hope you don't mind!). See https://help.github.com/articles/closing-issues-using-keywords/ for more info on that feature.

Thank you for the prompt attention and the new build!

PR #368, opened 2017-09-15
No longer throwing an exception when a tween is removed during update

iangilman:
Fixes https://github.com/tweenjs/tween.js/issues/367


iangilman:
Let me know if you want a unit test for this scenario.

Issue #375, opened 2017-09-16
`_Group` is too common , could you change it ?


finscn:
I see there is a  '_' prefix , but `_Group` is still too common.

Could you change it to   `_TweenGroup` ?

mikebolt:
Good call, leaked private variables should be uniquely named. I will merge
this when I have some time.

trusktr:
Before we merge, why does it matter that much? Nowadays, making simple names is a great thing, because they can be encapsulated in modules.

In this case, `Group` is currently housed in `TWEEN`, so we have `TWEEN.Group`. I think that's already clear, and that `TWEEN.TweenGroup` is now redundant because it is obvious what `TWEEN.Group` is.

When using a module system, users know what they are importing, and they can name it what they want. Here's examples using CommonJS and ES2015 modules:

```js
// CommonJS
const { Group: TweenGroup } = require('tween.js')
import { Group as TweenGroup} from 'tween.js'

// use TweenGroup
```

I like this approach more. What are your thoughts on it?

finscn:
The key point is not the name of `_Group` , current version tween.js ,  the `_Group` pollutes the global scope in browser. it's not a good sign.

So. I think the PR #377 is better.

trusktr:
👍 we should not pollute, let's get that fixed.

trusktr:
Related: #379

PR #376, opened 2017-09-16
Change `_Group` to `_TweenGroup`

finscn:
See #375


finscn:
I think  #377 is a better PR .

The name is not key point , the key point is "It's should not be public".

PR #377, opened 2017-09-17
Make `_Group` be a private class.

finscn:
Move `_Group` to a anonymous function , make it be private.
The purpose  of this PR is like #376 , but I think this one is better.

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -91,7 +91,6 @@ var TWEEN;
 		return TWEEN._nextId++;
 	};
 
-
 	// Include a performance.now polyfill.
 	// In node.js, use process.hrtime.
 	if (typeof (window) === 'undefined' && typeof (process) !== 'undefined') {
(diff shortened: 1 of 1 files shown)


finscn:
The comparing view looks very complex , but in fact , I just move some code into a `(function () { }`.

I think the Comparison Algorithm of git is not smart enough.

trusktr:
It's a good idea not to pollute globals. Idea: to clean up code, rather than having two variables (Group and _Group), let's just have one.

trusktr:
As an alternative to this PR, I made https://github.com/tweenjs/tween.js/pull/379 which simply moves everything into the UMD definition as it should've been in the first place (if we're talking about purely technical reasons, because if we're talking about aesthetics then, sure, the code is nicer to look at when it isn't indented inside the factory function).

This work will only be temporary anyways, we should get a build system in place ( #331 ).

src/Tween.js
```diff
@@ -11,12 +11,12 @@ var TWEEN;
 
 (function () {
 
-	var _Group = function () {
+	var Group = function () {
```
trusktr: 👍 

Thoughts on #379?

I think I'd prefer to have a single factory function. If you'd like to modify this PR to do similar to my PR, let's get it merged so you can get the credit. 😄

trusktr: @finscn Also note that the commit messages should follow semantic-release format. See here: https://github.com/semantic-release/semantic-release#default-commit-message-format

trusktr:
If you can also squash your commits into one, that'd be great. I don't mind multiple commits, but in this case a commit message like "Cosmetic" isn't too meaningful, so I think in this case all your changes can fit nicely in one commit.

PR #379, opened 2017-09-17
fix: avoid polluting globals, and as a result we also don't need underscore prefixes

trusktr:
This moves the creation of TWEEN inside the UMD module definition so that we can avoid polluting globals. As a result, we don't need to use underscore naming, so the code is cleaner.

This is a precursory step before having an actual module system ( PR https://github.com/tweenjs/tween.js/pull/331 by @martikaljuve ). If we can get that merged, the changes here won't be necessary. Or we can merge this to have it until 331 can be merged later.

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -7,906 +7,923 @@
  * Thank you all, you're awesome!
  */
 
-
-var _Group = function () {
-	this._tweens = {};
-	this._tweensAddedDuringUpdate = {};
-};
-
-_Group.prototype = {
-	getAll: function () {
-
-		return Object.keys(this._tweens).map(function (tweenId) {
-			return this._tweens[tweenId];
-		}.bind(this));
-
-	},
-
-	removeAll: function () {
-
-		this._tweens = {};
-
-	},
-
-	add: function (tween) {
-
-		this._tweens[tween.getId()] = tween;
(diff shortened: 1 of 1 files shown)


finscn:
When this PR would be merged ?

LukeShu:
This commit changed the indentation style of the main code from tabs to 4-spaces, which is then inconsistent with the UMD function, which is tabs.

You commented that you didn't like it in the comments on #377, but having "actual changes" and "indentation changes" as 2 separate commits seems a little silly, but it
 * makes it easier to review exactly what changed, and
 * makes rebasing easier; just rebase the "actual changes" commit, and then trivially re-create the "indentation changes" commit (which is relevant, as this PR no longer applies to master).

src/Tween.js
```diff
@@ -7,906 +7,923 @@
...
+		});
 
-		return Object.keys(this._tweens).map(function (tweenId) {
-			return this._tweens[tweenId];
-		}.bind(this));
+	} else if (typeof module !== 'undefined' && typeof exports === 'object') {
 
-	},
+		// Node.js
+		module.exports = factory();
```
LukeShu: Note that `module.exports` is a Node.js thing; in the CommonJS Modules/1.1 spec, it's named `exports`.  Many modules are both Node.js and CommonJS modules by setting `exports = module.exports = ...`

PR #386, opened 2017-10-04
Fix infinite recursion in readme example (#385)

### LINKED ISSUE #385: Infinite recursion loop in readme example causing performance degradation
### Version
16.6.0

### Steps to reproduce
Run the example code in any arbitrary setInterval loop.

### What is expected?
The recursion should stop once the tween is done, which could be easily done with an `.onComplete` function.

### What is actually happening?
The recursion never stops. In instances where this is used a lot (i.e. multiple real-time update charts) this leads to some very quick performance degradation. 

### Demonstration
https://johnnynotsolucky.github.io/samples/vuejs-d3/
This runs the example code on a data set every 2s, so leave it running for a while (up to 3 minutes) to observe the performance degradation. Looking at the performance tab in firefox, you'll see *a lot* of `requestAnimationFrame` calls being fired as a result.

<br>

**Note**
I'm aware that this may have been left out of the example for the sake of simplicity, but realistically, most people will not look beyond the first examples, and it gets even worse when that code is used in other tutorials, where users will likely not look at the docs at all.

Edit: I made a [PR concerning this issue](https://github.com/tweenjs/tween.js/pull/386).

Kaptard:
Stops the recursive loop once the tween is done. Note that the codepen example will also require an update.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,13 +15,6 @@ box.style.setProperty('width', '100px');
 box.style.setProperty('height', '100px');
 document.body.appendChild(box);
 
-// Setup the animation loop.
-function animate(time) {
-    requestAnimationFrame(animate);
-    TWEEN.update(time);
-}
-requestAnimationFrame(animate);
-
 var coords = { x: 0, y: 0 }; // Start at (0, 0)
 var tween = new TWEEN.Tween(coords) // Create a new tween that modifies 'coords'.
         .to({ x: 300, y: 200 }, 1000) // Move to (300, 200) in 1 second.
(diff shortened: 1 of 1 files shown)


Kaptard:
Note that the codepen example will also require an update with the new code.

lokielse:
@Kaptard Thanks, it works!

lokielse:
```js
let animate = function(time) {
    if (TWEEN.update(time)) {
       requestAnimationFrame(animate)
    }
}

let tween = new TWEEN.Tween(...)
...
tween.start()

requestAnimationFrame(animate)
```

mikebolt:
I would prefer to keep the demo example as simple as possible, and I also think that the normal use case requires keeping the RAF loop running in case more tweens are started as the result of a user interaction. Also it's not a performance issue because RAF will be called at most 60 or 90 times per second.

If this is causing a performance issue or a memory leak please open a new issue.

PR #388, opened 2017-10-25
Fix the examples.

mikebolt:
Finally!

### DIFF
--- a/examples/01_bars.html
+++ b/examples/01_bars.html
@@ -48,8 +48,8 @@
 
 					var elem = { x: startValue, domElement: domElement };
 
-					var updateCallback = function() {
-						this.domElement.style.left = this.x + 'px';
+					var updateCallback = function(object) {
+						object.domElement.style.left = this.x + 'px';
 					}
 
 					var tween = new TWEEN.Tween(elem)
--- a/examples/01_bars.html
+++ b/examples/01_bars.html
@@ -49,7 +49,7 @@
 					var elem = { x: startValue, domElement: domElement };
 
 					var updateCallback = function(object) {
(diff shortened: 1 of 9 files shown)


trusktr:
Changes LGTM. Anything else or is that all that is needed?

mikebolt:
That’s it!

examples/01_bars.html
```diff
@@ -48,8 +48,8 @@
 
 					var elem = { x: startValue, domElement: domElement };
 
-					var updateCallback = function() {
-						this.domElement.style.left = this.x + 'px';
+					var updateCallback = function(object) {
+						object.domElement.style.left = this.x + 'px';
```
martikaljuve: `this.x` needs to be changed to `object.x`.

examples/05_video_and_time.html
```diff
@@ -40,9 +40,9 @@
 					new TWEEN.Tween( { x: 50, y: 0 } )
 						.to( { x: 400 }, this.duration )
```
martikaljuve: I'm getting `NaN` for `this.duration` when testing this example locally. The video metadata might not be loaded yet when the 'play' event is triggered. Maybe we should use 'playing' event instead.

examples/01_bars.html
mikebolt: Fixed

examples/05_video_and_time.html
mikebolt: I don't have this problem. Could you create a separate issue for this? I would like to get this fix merged.

trusktr:
In general, should we add tests for the examples?

PR #394, opened 2017-10-29
Zero duration

### LINKED ISSUE #384: NaN elapsed percentage when start time is equal to update time and duration is zero.
In the following code https://github.com/tweenjs/tween.js/blob/678cd11a909990b3ff007d458008843939626b37/src/Tween.js#L344

```
elapsed = (time - this._startTime) / this._duration;
elapsed = elapsed > 1 ? 1 : elapsed;
```

elapsed can end up as `NaN` when time === this._startTime and this._duration === 0.

This is explained here: https://stackoverflow.com/questions/18838301/in-javascript-why-does-zero-divided-by-zero-return-nan-but-any-other-divided-b

(TLDR; `1/0 === Infinity` and `1/0 > 1 === true`  vs `isNaN(0/0) === true` and `0/0 > 1 === false`)

The fix for this is to change the code to:

```
elapsed = this._duration === 0 ? 1 : (time - this._startTime) / this._duration;
elapsed = elapsed > 1 ? 1 : elapsed;
```

mikebolt:
Fixes #384 

This adds a single boolean test to the hot-path.

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -342,7 +342,7 @@ TWEEN.Tween.prototype = {
 		}
 
 		elapsed = (time - this._startTime) / this._duration;
-		elapsed = elapsed > 1 ? 1 : elapsed;
+		elapsed = (this._duration === 0 || elapsed > 1) ? 1 : elapsed;
 
 		value = this._easingFunction(elapsed);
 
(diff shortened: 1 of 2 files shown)


mikebolt:
Let's merge this. I want to get some PRs merged and do a new release.

Issue #399, opened 2017-12-08
Breaks in Safari 10


Sailias:
https://bugs.webkit.org/show_bug.cgi?id=171041

The following functions break in older versions of Safari before they patched the bug above:

```
        yoyo: function yoyo(yoyo) {

		this._yoyo = yoyo;
		return this;

	},

	easing: function easing(easing) {

		this._easingFunction = easing;
		return this;

	},

	interpolation: function interpolation(interpolation) {

		this._interpolationFunction = interpolation;
		return this;

	},
```

PR #400, opened 2017-12-08
refs #399 - Supporting older versions of safari

Sailias:


mikebolt:
Is this problem fixed if we use anonymous functions instead?

trusktr:
Anonymous functions don't debug as nicely in some debuggers. I think this fix is better, though the code is only slightly uglier.

mikebolt:
thanks

PR #401, opened 2017-12-27
Replace 'this' in all examples

BoTime:
Also fixed a issue in "01_bars.html" potentially caused by closure.

### DIFF
--- a/examples/01_bars.html
+++ b/examples/01_bars.html
@@ -48,21 +48,21 @@
 
 					var elem = { x: startValue, domElement: domElement };
 
-					var updateCallback = function() {
-						this.domElement.style.left = this.x + 'px';
+					var updateCallback = function(elem) {
+						elem.domElement.style.left = elem.x + 'px';
 					}
 
 					var tween = new TWEEN.Tween(elem)
 						.to({ x: endValue }, 4000)
 						.delay(Math.random() * 1000)
-						.onUpdate(updateCallback)
+						.onUpdate(updateCallback.bind(null, elem))
 						.easing(TWEEN.Easing.Back.Out)
 						.start();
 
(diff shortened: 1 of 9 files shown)


examples/04_simplest.html
```diff
@@ -24,14 +24,13 @@ <h2>04 _ simplest possible example</h2>
 				var output = document.createElement( 'div' );
 				output.style.cssText = 'position: absolute; left: 50px; top: 300px; font-size: 100px';
 				document.body.appendChild( output );
-
+				var startValue = { x: 50, y: 0 };
```
trusktr: @BoTime Thanks for making this effort. Have you tested this example? Looks like you need to pass startValue into the new Tween, otherwise the value output is constant.

examples/01_bars.html
```diff
@@ -48,21 +48,21 @@ <h2>01 _ Bars</h2>
 
 					var elem = { x: startValue, domElement: domElement };
 
-					var updateCallback = function() {
-						this.domElement.style.left = this.x + 'px';
+					var updateCallback = function(elem) {
+						elem.domElement.style.left = elem.x + 'px';
```
trusktr: Don't think you need the `elem` arg here, and also no `.bind` because the `var elem` is in scope, that'd be cleaner.

examples/05_video_and_time.html
```diff
@@ -36,7 +36,7 @@ <h2>05 _ video and time</h2>
 
 				video = document.getElementById( 'video' );
 				video.addEventListener( 'play', function() {
-					
+
```
trusktr: Missing some `this` in this file. Are there any more files that need `this` changed?

martikaljuve:
PR #388 also fixes the examples.

PR #403, opened 2018-01-18
changed named functions to anonymous

OlegSavvateev:
```js
yoyo: function yoyo(yoyo) {

	this._yoyo = yoyo;
	return this;

},
```
This code does not work with old safari browsers, because a function has same name as a parameter. This pull request fix it.


mikebolt:
This was fixed in #399, which I just merged. Could you test out the current version of the code in Safari?

trusktr:
This PR modifies some other functions additional from #399. @OlegSavvateev It'd be great if you can rebase on master and see which functions still need fixing.

PR #408, opened 2018-02-01
create a Chinese(zh-CN) version

zhaoqize:
tween.js is a great library.
Due to the use of the Chinese version is not convenient,So I decided to translate tween.js.


trusktr:
This is awesome, but can we count on you to update the Chinese version if we update the README?

zhaoqize:
@trusktr of course! I will continue to pay attention to it.If you accept, this is my pleasure.

trusktr:
@zhaoqize Alright. @mikebolt Thoughts?

PR #423, opened 2018-08-31
Added multiplication and division operations to relative values

upisfree:


upisfree:
Need to finalize it: I saw some bugs.

PR #427, opened 2018-09-24
user_guide.md - fixed typos and tweaked intro

rtpHarry:
spotted a spelling mistake, then fixed a few other issues, then tweaked the intro as there aren't any gaps or TODO sections in the docs any more.

I dropped it into Grammarly just to double check for any other issues and it seems to have messed with the code indentation but it still displays ok so I'm not going to redo the whole thing unless you request it.


HungryZombie:
4AEE18F83AFDEB23

PR #438, opened 2019-01-28
Adds web worker capability

bodo22:
By replacing the usage of `window` with `self`, it's possible to use this lib in web workers, e.g. very useful when rendering offscreen canvases in web workers.

https://developer.mozilla.org/en-US/docs/Web/API/Window/self

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -91,7 +91,7 @@ TWEEN.nextId = function () {
 
 // Include a performance.now polyfill.
 // In node.js, use process.hrtime.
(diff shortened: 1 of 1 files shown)


bodo22:
@mikebolt can i help make this PR mergable, do you see any concerns with the replacements?

mikebolt:
I was waiting to test this in node.js and I got lazy.

I haven't actually tested this in a web worker. If you added a web worker example that would be cool.

bodo22:
Thanks for merging!

There's currently a [chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=935096), breaking potential example test cases. I'll try and find a workaround in the mean time.

PR #441, opened 2019-02-19
+Make tween working with simple `number`.

paladin-t:
To allow using numbers as begin/complete values, i.e.

```js
var tween = new TWEEN.Tween(0)
	.to(100, 1000)
	.easing(TWEEN.Easing.Quadratic.Out)
	.onUpdate(function(v) {
		console.log(v);
	})
	.start();
```

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -165,7 +165,7 @@ TWEEN.Tween.prototype = {
 		}
 
 		if (this._object.boxed)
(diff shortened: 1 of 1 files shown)


mikebolt:
I'm really sorry but I'm not going to merge this. The user can tween single number values by making a `{value: x}` object wrapper then access it in the callbacks with `.value`. This PR is effectively syntactic sugar for that code, and it comes at the cost of some complexity and an extra branch in every update callback.

PR #415, opened 2018-05-21
Remove faulty semicolon from code in readme

weiserhei:
This semicolon prevents this code from working for me in Firefox, you might want to remove it.

PR #463, opened 2019-08-19
Remove faulty semicolon from user_guide

weiserhei:
Removing the faulty semicolon again, as this code just doesnt work with it. Users copy-n-paste this code and have to debug :crying_cat_face: 

See https://github.com/tweenjs/tween.js/pull/415 and this PR which re-added the semicolon: https://github.com/tweenjs/tween.js/pull/427


mikebolt:
Sorry about that, and thanks for the PR. Looks like it got ignored in a merge.

PR #483, opened 2019-09-10
Fix Tween.js path in examples.

chepe263:
I also changed some `<title /> `to make them homogeneous with the rest of the examples.


mikebolt:
Thanks. The gh-pages branch is really old. I need to set up TravisCI to push to the gh-pages branch when it deploys.

Issue #341, opened 2017-06-19
Add a pause method


mikebolt:
The title says it all.

This can be used for a reference: https://github.com/CasualBot/tween.js

gigablox:
This is great! Using your lib for now until this gets merged in!

brannondorsey:
This is a feature that would be very useful in `master`.

alx696:
#15 #44 #135 从2011年提出到现在,还是没有添加. 这个功能非常有用, 比如轨迹播放. 经过了一番尝试,下面的方法有效. 已在ThreeJS的相机轨迹控制中验证.

总共需要4步(4 Step):
1. 定义全局变量,用于缓存信息;
2. 在requestAnimationFrame调用函数中添加控制逻辑(通过TWEEN.update()修改当前时间);
3. Tween执行完毕后重置全局变量;
4. 设置暂停/恢复标记.

>  var `S` is `this`.

### 1. define var
```
  tweenPauseFlag: false, //Tween是否暂停
  tweenPauseLastAnimateTime: 0, //Tween暂停后上一次Animate时间
  tweenPauseTimeSum: 0, //Tween暂停合计时间(用于将Tween的时间轴平行后移)
```

### 2. change tween time
```
    function animate(time) {
      requestAnimationFrame(animate);

      //处理Tween暂停
      if (S.tweenPauseFlag) {
        if (S.tweenPauseLastAnimateTime === 0) {
          //检测到暂停,缓存当前time
          S.tweenPauseLastAnimateTime = time;
          console.debug('Tween暂停(animate time):', S.tweenPauseLastAnimateTime);
        } else {
          //计算暂停的时长并累加
          S.tweenPauseTimeSum += time - S.tweenPauseLastAnimateTime;
          //重新缓存当前time
          S.tweenPauseLastAnimateTime = time;
        }
      } else if (S.tweenPauseLastAnimateTime !== 0) {
        //重置当前time缓存
        S.tweenPauseLastAnimateTime = 0;
        console.debug('Tween恢复');
      }
      TWEEN.update(time - S.tweenPauseTimeSum); //用当前time减去暂停时长,保持Tween时间轴不动.
    }

    requestAnimationFrame(animate);
```

### 3. rest var
When the current tween complete, rest var. If not, next tween will pause before start.
```
    //重置Tween暂停计时
    S.tweenPauseTimeSum = 0;
    S.tweenPauseLastAnimateTime = 0;
```

### 4. set flag
[...]

PR #484, opened 2019-09-10
Add functions to pause and resume a tween.

chepe263:
For issue #341 

I used https://github.com/CasualBot/tween.js as a reference/inspiration.

### DIFF
--- a/examples/14.pause_tween.html
+++ b/examples/14.pause_tween.html
@@ -1,0 +1,82 @@
+<!DOCTYPE html> 
+<html lang="en">
+	<head>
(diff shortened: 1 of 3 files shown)


mikebolt:
This won't work correctly when TWEEN.now() is not used as the time scale. `pause` and `resume` should accept a time parameter, like `update` does. It should default to TWEEN.now() if the parameter is not provided.

mikebolt:
Thanks for the PR.

chepe263:
Thank you for correcting my code.

PR #509, opened 2019-12-09
Replace const with var

mikebolt:
The "export const" statement causes Rollup to write "const" in the builds. Using "export var" instead prevents this.


trusktr:
Hello! :wave:  This is an interesting change. Why change it to `var`?

PR #366, opened 2017-08-30
Tween nested object

talltyler:
There didn't seem to be enough love for https://github.com/tweenjs/tween.js/pull/313 to get it merged in. To move the ball forward I made a version based on the suggestion by @mikebolt to make this work without the string interpolation.

The performance impact should be a single function call for each level of nesting.

PR #520, opened 2020-05-18
Handling nested properties

Malows:
Now you can tween nested properties.
Also, there's a small tweak of performance using JSON.parse for copying the end values object instead referencing the prototype ([more info](https://v8.dev/blog/cost-of-javascript-2019#json)).
And the performance impact should be the same as the original version, unless you're passing deep nested objects. Otherwise it's a single function call for each level of nesting.


trusktr:
Hello! Thanks for making a PR!

We have a new Pull Request tracking board at https://github.com/tweenjs/tween.js/projects/4, I added the PR there. Looking to make more progress with Tween.js.

trusktr:
I have a request: can you please separate this into two PRs: first one that makes the change, then a second that can refactor the code into separate functions. It is easier to review (and merge with other ideas) that way.

Malows:
Hi! Yes, the PR it's based in the #366 suggestions, but I thought it was a good idea avoid the prototype sharing and using TDD in the process to ensure not breaking any other features. Let me move the helpers functions inside the Tween definition

src/Tween.js
```diff
@@ -164,7 +164,7 @@ TWEEN.Tween.prototype = {
 
 	to: function (properties, duration) {
 
-		this._valuesEnd = Object.create(properties);
+		this._valuesEnd = JSON.parse(JSON.stringify(properties));
```
trusktr, line 167: This is only faster when using object literals, as in `Object.create({"foo": 123, "bar": 123})` vs `JSON.parse(JSON.stringify('{"foo": 123, "bar": 123}'))`.

Otherwise if we're dealing with references JSON.parse/stringify will be slower because there's nothing to parse if we're passing a reference to `Object.create`.

Performance test: https://jsperf.com/object-create-o-vs-json-parse-json-stringify-o/

trusktr:
I am making some updates, and I will re-open this on a new PR. Thanks!

trusktr:
moved to https://github.com/tweenjs/tween.js/pull/536

Issue #284, opened 2016-08-22
yoyo doesn't work for Tweening of Arrays


goldenratio:
Sample code:

yoyo doesn't work for Tweening of Arrays

```
 new TWEEN.Tween({posX: 100, posY: 180})
                .to({posX: [320, 570, 800], posY: [140, 260, 160]}, duration)
                .easing(ease)
                .yoyo(true)
                .repeat(Infinity)
                .delay(delay)
                .interpolation(TWEEN.Interpolation.Linear)
                .onUpdate(function()
                {
                    if(thisObject._ball)
                    {
                        thisObject._ball.position.set(this.posX, this.posY);
                    }

                })
                .start();
```

mikebolt:
I found the problem and commited a fix for the issue to a branch in my fork. [Here is the fixed Tween.js file](https://github.com/mikebolt/tween.js/blob/array-yoyo-fix/src/Tween.js). Unfortunately the array tweening functionality didn't have any unit tests beforehand, so I need to add some tests before merging it into master.

sole:
If anyone wants to chime in and help, we need tests for array tweening. Then we can merge the PR (although it might need some 'rebasing' work since it's been a while since the branch was created).

Tests should be added to the [test/unit/tests.js file](https://github.com/tweenjs/tween.js/blob/master/test/unit/tests.js)

dalisoft:
Maybe we do `reverse/yoyo` backward-elapsed-time for make it work? I have already working way, but i think it looks like self-promoting (in other js, based-on Tween.js)

jshawl:
Added unit tests for @mikebolt's solution!

trusktr:
Is there a PR for this?

mikebolt:
I can make a PR from my fork but we still need tests for array tweens.

I actually don't remember how I fixed this 😃

trusktr:
@mikebolt That'd be great if you can to get it started, then at least one of us could add tests to it, etc. 😄

trusktr:
pull request opened for this at https://github.com/tweenjs/tween.js/pull/523

trusktr:
Oh nevermind, https://github.com/tweenjs/tween.js/pull/320 was already opened for that.

PR #523, opened 2020-05-30
Fix issue with yoyo'd array value tweens.

trusktr:
fixes https://github.com/tweenjs/tween.js/issues/284


trusktr:
Closed because there was already another PR with the changes.

PR #526, opened 2020-05-30
Fix to make yoyo work with relative values

### LINKED ISSUE #178: Yoyo not working properly when relative values are used
If we use relative values in a tween and yoyo() along with it, then its not working as expected. Following code shows the error:

```
var position = {x:0, y:150};
        var tween0 = new TWEEN.Tween(position)
                .to({x: "+50"}, 1000)
                .repeat(1)
                .yoyo(true)
                .onComplete(function() {
                    console.log("Tween 0 completed. X is " + this.x);
                })
                .onUpdate(function() {
                    console.log("x:" + this.x);
                });
        tween0.start();
```

Here the value of 'x' should be between 0 and 50 (both included), but its going out of this range, like values 500.

trusktr:
Fixes #178


trusktr:
Added tests and updated the examples. Thanks @sinujohn!

PR #527, opened 2020-05-30
support relative values in arrays passed to the .to() method

trusktr:


.editorconfig
```diff
@@ -0,0 +1,15 @@
...
+[*]
+charset = utf-8
+indent_style = tab
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+indent_size = 2
+indent_style = space
```
trusktr, line 15: My editor was wonky without this, so I decided to finally add it. It's very nifty.

PR #511, opened 2019-12-26
bug-fix: update function still be called after end();

Nafeng:

PR #529, opened 2020-05-30
bug-fix: update function still be called after end();

trusktr:
moved from https://github.com/tweenjs/tween.js/pull/511


trusktr:
@Nafeng Now that I got tests back up and running, they aren't passing yet. This is currently interfering with infinite repeat tweens.

PR #531, opened 2020-05-30
Timescale

trusktr:


trusktr:
I think we'll close this, and someone can scale outside of the tween with more likelihood of not breaking internals. For example, just change the duration of a tween from the outside to get the desired effect.

PR #320, opened 2017-02-16
Allow tween to array with yoyo

jshawl:
fixes https://github.com/tweenjs/tween.js/issues/284

PR #536, opened 2020-05-31
Nested properties

trusktr:
closes https://github.com/tweenjs/tween.js/issues/260


trusktr:
@Malows I merged all the latest from master, fixed conflicts. The complex-object tests are passing, but for some reason the `yoyo works with arrays` test isn't passing.

I changed over to using `for (var prop in foo)` instead of `Object.create` (instead of `JSON.parse`). because the `Object.create` wasn't working when it came to some one of the fixes involving `yoyo`.

Also there are no tests for `repeat` and `yoyo` with regards to complex objects yet.

trusktr:
Alright, tests are passing (the error was with https://github.com/tweenjs/tween.js/pull/320 which I merged yesterday, and this PR was not yet handling that case).

Things to do:

- [ ] tests for nested values with repeat
- [ ] tests for nested values with yoyo
- [ ] possibly clean up some logic (conditionals can be simplified, but moving anything around has high chance of exploding, not specifically with regards to nested values but with all that's been compounded over time)

trusktr:
I think I'll just merge this. As tests are passing, nothing previous is broken, we just need to verify the repeat/yoyo functionality next.

PR #541, opened 2020-05-31
Test animated values in nested array

trusktr:
Added support for values in nested arrays (adding on top of @Malows's and @talltyler's nested object support)

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -223,7 +223,16 @@ TWEEN.Tween.prototype = {
 
 		for (var property in _valuesEnd) {
 
-			var isInterpolationList = Array.isArray(_valuesEnd[property]);
+			var startValue = _object[property];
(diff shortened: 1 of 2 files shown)


src/Tween.js
```diff
@@ -566,12 +566,15 @@ TWEEN.Tween.prototype = {
...
+			var endIsArray = Array.isArray(end);
+			var isInterpolationList = !startIsArray && endIsArray;
 
-			if (Array.isArray(end)) {
+			if (isInterpolationList) {
 
 				_object[property] = this._interpolationFunction(end, value);
 
-			} else if (end && typeof end === 'object') {
+			} else if (typeof end === 'object' && end) {
```
trusktr, line 577: bug fix, so it doesn't skip this condition on `0` values.

PR #510, opened 2019-12-15
Migration to Typescript maintaining backward compatibility.

nidorx:
Migration to Typescript **maintaining backward compatibility**.

1. Pass all original tests and examples without change
2. **eslint** - Uses the recommended rules from the `@typescript-eslint/eslint-plugin` (`@typescript-eslint/recommended`, `prettier/@typescript-eslint`, `prettier/recommended` )
3. Resolved dependency vulnerabilities:

**before**
```
$ npm i
audited 1045 packages in 3.265s
found 79 vulnerabilities (23 low, 20 moderate, 35 high, 1 critical)
```
**now**
```
$ npm i
audited 897 packages in 2.437s
found 0 vulnerabilities
```

PR #544, opened 2020-06-03
Convert to TypeScript

trusktr:
moved from https://github.com/tweenjs/tween.js/pull/510


trusktr:
Done! Thanks @nidorx!

amelnikoff:
Hello.
Can you provide some examples to readme.md how to use TS version correctly?

trusktr:
@amelnikoff Hello! It is fairly self explanatory if you know how to import TS libs into a TS project. `npm install`, then in your code `import {whatever} from '@tweenjs/tween.js'`.

We should add an example though. I think an `example` folder will be nice, with examples that are self-contained working little apps (f.e. importing with bundled ES modules in JS vs TS using build tools, importing as a global script tag, or importing as native ES Modules, etc).

amelnikoff:
@trusktr thanks. Thanks, simple `import {whatever} from '@tweenjs/tween.js'` is enough for me and I think for most developers ;)

Issue #545, opened 2020-06-03
fix the 07_dynamic_to example.


trusktr:
The 07_dynamic_to example broke recently (My guess is it might be due to the new nested object feature).

I will fix it and add tests for it. I've been adding tests along with each PR, but that particular feature has no tests in place.

**This needs to be fixed before the next release so as not to introduce a breaking change.**

I will merge the TypeScript pull request, as that has no effect (the example is broken the same before and after, but otherwise all tests and example pass in the TS PR).

trusktr:
I will bisect this soon...

mikebolt:
This is part of the problem: https://github.com/tweenjs/tween.js/blob/master/src/Tween.ts#L57

It was added here: https://github.com/tweenjs/tween.js/commit/f82ee2598ee79f7cc092e37f6fabe11dc5d5ef6f

trusktr:
Awesome. Thank you for looking into this. What problem did you see there?

mikebolt:
Object.create makes a prototypal copy, not an actual copy. Therefore when the object passed to `.to()` is updated, its updated properties are visible on _valuesEnd when it's created with Object.create, but tween.js can't modify the actual object that was passed in (without using `.__proto`).

trusktr:
Thanks! I'm planning to circle back around to this soon.

trusktr:
PR: https://github.com/tweenjs/tween.js/pull/578

trusktr:
released in 18.6.3

PR #549, opened 2020-06-21
18.6.0

mikebolt:
bumping version


trusktr:
I meant for https://github.com/tweenjs/tween.js/issues/545 to be fixed before releasing an in-range breaking change.

What protocol should we follow so this doesn't happen again?

trusktr:
Ah, I have an idea: perhaps we can add a prepublish hook that causes npm publish to fail if there are any open issues tagged with `required next release`. Thoughts on that?

Issue #417, opened 2018-06-05
delay is multiplied by ten when start() is given a ms value as a string


EmNudge:
It isn't written in the docs that start() even can get passed a value instead of using delay(), but it can. What's weird is that, when passed a string instead of a number, the number is parsed and then multiplied by ten.
**start(5000)** gives it a 5 second delay, while **start('5000')** gives it a 50 second delay.

If I had to guess what the problem is, I'd say it's probably by
`		this._startTime = time !== undefined ? typeof time === 'string' ? TWEEN.now() + parseFloat(time) : time : TWEEN.now();
		this._startTime += this._delayTime;`
and when it's adding the delayTime, it's concatenating them (i.e. parseFloat() didn't work and this._startTime is registering as a string. Thus it added a 0 to itself and only works as a number later on).

fix? I guess something like `this._startTime = parseFloat(this._startTime) + this._delayTime` but I'm not really sure why the issue appears in the first place.

trusktr:
I fixed the relative_start_time example today. We should have been more careful when merging things, to keep other things up to date.

PR #550, opened 2020-06-21
remove support for string arguments to start and mark the argument as…

mikebolt:
… optional

### DIFF
--- a/src/Tween.ts
+++ b/src/Tween.ts
@@ -70,7 +70,7 @@ export class Tween<T extends UnknownProps> {
 		return this
 	}
 
-	start(time: number): this {
+	start(time?: number): this {
 		if (this._isPlaying) {
 			return this
 		}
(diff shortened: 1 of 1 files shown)


mikebolt:
see #417

src/Tween.ts
```diff
@@ -70,7 +70,7 @@ export class Tween<T extends UnknownProps> {
 		return this
 	}
 
-	start(time: number): this {
+	start(time?: number): this {
```
trusktr, line 73: :ok

src/Tween.ts
```diff
@@ -101,8 +101,13 @@ export class Tween<T extends UnknownProps> {
 
 		this._isChainStopped = false
 
-		this._startTime =
-			time !== undefined ? (typeof time === 'string' ? TWEEN.now() + parseFloat(time) : time) : TWEEN.now()
+		if (time === undefined) {
+			this._startTime = TWEEN.now()
+		}
+		else {
+			this._startTime = time
```
trusktr, line 108: What about `TWEEN.now() + time`?

trusktr:
`npm run lint` will fix the formatting and the tests will pass. But I'm curious if removing the `now()+time` is a breaking change.

src/Tween.ts
trusktr: On second though, passing a string looks totally broken here. I'll fix this up and remove.

PR #558, opened 2020-07-22
Typing: time parameter in update() and start() is now optional

niconoe:
Hi, 

I made the `time` parameter optional, to reflect the examples and documentation. For `update()`, the TS compiler warned me of a possibly undefined value, so I added a line of code to make sure the current time is used if none is provided. This is consistent to what I read in the documentation and the tests all pass, but please review this change critically since actually don't know much about Tween.js.

### DIFF
--- a/src/Tween.ts
+++ b/src/Tween.ts
@@ -70,7 +70,7 @@ export class Tween<T extends UnknownProps> {
 		return this
 	}
 
-	start(time: number): this {
+	start(time?: number): this {
 		if (this._isPlaying) {
 			return this
 		}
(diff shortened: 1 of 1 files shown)


trusktr:
Hello @niconoe Thanks for the change! I merged it, but it would be great to add a test for `update` without any time value.

mikebolt:
It's potentially difficult to test because the current time will change
depending on when the test is run, and the tests need to run in both node
and browsers. You could copy the "now" polyfill that we made. Also, the
test might be flaky due to the clock advancing by 1ms, so give it a buffer
of 5ms or something.

Issue #557, opened 2020-07-06
When the browser tab sleeps, the delay of repeated tweens is not keep


MasatoMakino:
## Description

When the browser tab goes into the background, the delay of repeated tween is not keep.

## Expected Behavior

Tabs are restored from the background, the delay of tween is unchanged.

## Actual Behavior

Tabs are restored from the background, the delay of tween changes.

## Steps to Reproduce

1. open https://codepen.io/masatomakino/pen/QWyQpYY
2. open new tab.
3. change the tab of codepen to the background and wait 5 seconds. ( = ``tween._repeatDelayTime``)
4. change the codepen tab to foreground.
5. delay of tween is not kept. 

## Possible Fix

When the browser tabs are in the background, ``requestAnimationFrame`` will go to sleep. requestAnimationFrame will return, and the pass large delta time.

However, ``tween.js`` loop does not take large delta time.

```js
elapsed = this._duration === 0 || elapsed > 1 ? 1 : elapsed
```
[source code](https://github.com/tweenjs/tween.js/blob/e9a59f1fcd33d3a31ba99b350fdc37b1ba86c9c0/src/Tween.ts#L349)

In this process, time is rounded to ``elapsed``. And the delta time disappears.

````js
const overflowCounter = Math.trunc((time - this._startTime) / this._duration);
const overflowTime = (time - this._startTime) % this._duration;
````
For example, we get an overflow for this process. You can use this value to modify ``this._repeat `` and ``this._startTime``.

[source code](https://github.com/tweenjs/tween.js/blob/e9a59f1fcd33d3a31ba99b350fdc37b1ba86c9c0/src/Tween.ts#L386)

## Environment

macOS 10.15.2
Chrome 83.0.4103.116 (Official Build)
tween.js 18.6.0

Thank you.

trusktr:
Hello! Yeah, this is a known issue. The Tween code merely gets the current time, it does not account for time that the window is unfocused and `requestAnimationFrame` can not run.

The current workaround is to manually update tweens, passing in our own time values to `tween.update()`, and then we can manage the time any way we want.

It would be great to fix this. If interested in making a PR, happy to review.

trusktr:
.repeat is deprecated in 

- #713 

and will eventually be removed to cleanup Tween and reduce bugs from too much complexity intertwined in one spot.

`Timeline` is the new way to orchestrate.

```js
// "repeat" twice
timeline
  .add(tween)
  .add(tween)

[...]

PR #564, opened 2020-10-03
Handling of repeats with browser tab sleep

MasatoMakino:
## Related Issues

#557
When the browser tab sleeps, the delay of repeated tweens is not keep

This fixes handling of repeats.

## Details of the changes

When interval for TWEEN.update() is more than one repeat

+ Continuous elapsed
+ Multiple `tween._repeat` subtractions


trusktr:
Thanks for doing this! Will review it soon.

src/Tween.ts
```diff
@@ -348,9 +347,28 @@ export class Tween<T extends UnknownProps> {
...
+		const totalTime = this._duration + this._repeat * durationAndDelay
 
+		const calculateElapsedPortion = () => {
+			if (this._duration === 0) return 1
+			if (elapsedTime > totalTime) {
+				return 1
+			}
+
+			const timesRepeated = Math.trunc(elapsedTime / durationAndDelay)
+			const timeIntoCurrentRepeat = elapsedTime - timesRepeated * durationAndDelay
```
trusktr, line 361: @MasatoMakino Hello! I renamed some vars for easier reading. F.e. I renamed `elapsedTime` to `timeIntoCurrentRepeat`.

MasatoMakino: I agree with what @trusktr said.

trusktr: I feel like the naming isn't quite right yet. 🤔

src/Tween.ts
```diff
@@ -386,11 +405,7 @@ export class Tween<T extends UnknownProps> {
 					this._reversed = !this._reversed
 				}
 
-				if (this._repeatDelayTime !== undefined) {
-					this._startTime = time + this._repeatDelayTime
-				} else {
-					this._startTime = time + this._delayTime
-				}
+				this._startTime += durationAndDelay * completeCount
```
trusktr, line 408: @MasatoMakino Interesting. Looks like this fixes an inaccuracy! Before we had `_startTime = time + this._repeatDelayTime`, which means that if the `time` is beyond the beginning of a repeat, then the next repeat will happen after `this._repeatDelayTime` plus a small value, but it should instead happen at exactly the right time so that the _next_ time is relative to the correct starting time.

Difficult to describe with words. Good find!

src/Tween.ts
```diff
@@ -360,10 +378,11 @@ export class Tween<T extends UnknownProps> {
 			this._onUpdateCallback(this._object, elapsed)
 		}
 
-		if (elapsed === 1) {
+		if (this._duration === 0 || elapsedTime >= this._duration) {
 			if (this._repeat > 0) {
+				const completeCount = Math.min(Math.trunc((elapsedTime - this._duration) / durationAndDelay) + 1, this._repeat)
 				if (isFinite(this._repeat)) {
-					this._repeat--
+					this._repeat -= completeCount
```
trusktr, line 385: [Down below](https://github.com/tweenjs/tween.js/pull/564/files#diff-6f8cd5c42458ee058628fa97ed5c5c4bf529a252a662dc150844217f273f2561R423), regarding the chained tweens, there is this:

```js
					this._chainedTweens[i].start(this._startTime + this._duration)
```

Does this need to be updated too? Should it now be only `this._chainedTweens[i].start(this._startTime)` because we already advanced _startTime to where the next repeat would have been?

trusktr: This chained tween logic happens on the _next_ update right? Should I wonder if it needs to happen immediately in the updated where _repeat > 0, f.e. when we run `this._repeat -= completeCount` and _repeat becomes 0.

trusktr: This is getting complicated. I think removing some complexity from here, and instead adding something like a Timeline feature would make this a lot easier to understand:

#647

trusktr: Another issue with repeat(N) is that we cannot make the time go backwards, because we lose the initial _startTime every time we increment it. Timeline would make it easy to go forwards/backward anywhere on the timeline.

src/Tween.ts
MasatoMakino: I appreciate the review!

src/Tween.ts
MasatoMakino: The use of "repeat" and "chain" together is complicated. I agree with the need for additional unit tests.

trusktr:
Merged! All existing tests pass, plus the new tests pass, so this is an improvement even if we know it is not working with all features.

What we should do now is start to prevent more complexity, and next we should focus on [simplifying](https://github.com/tweenjs/tween.js/issues/655), and then [Timeline](https://github.com/tweenjs/tween.js/issues/647).

trusktr:
Ah shoot, I'm doing some manual testing, and the `13_relative_start_time` example broke. Will revert.

trusktr:
Hmmm, well, the `13_relative_start_time` uses `.delay(-500)`, which is weird. I'm going to delete that use case from the example, which is not even documented, and keep the change from this PR which is more valuable, and mark it as a breaking change.

I literally have no idea what "delay negative 500" even means, no point in keeping confusing API.

trusktr:
Released in v23.0.0

PR #578, opened 2020-10-22
fix: get dynamic to working again (for simple objects), add tests

### LINKED ISSUE #545: fix the 07_dynamic_to example.

trusktr:
fixes #545

### DIFF
--- a/examples/07_dynamic_to.html
+++ b/examples/07_dynamic_to.html
@@ -47,6 +47,7 @@
 
 				new TWEEN.Tween(rabbit)
 					.to({x: width - 50, y: height - 50}, 3000)
+					.easing(TWEEN.Easing.Exponential.InOut)
 					.onUpdate(function (object) {
 						// draw background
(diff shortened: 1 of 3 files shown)


examples/07_dynamic_to.html
```diff
@@ -47,6 +47,7 @@ <h2>07 _ dynamic to</h2>
 
 				new TWEEN.Tween(rabbit)
 					.to({x: width - 50, y: height - 50}, 3000)
+					.easing(TWEEN.Easing.Exponential.InOut)
```
trusktr, line 50: looks prettier :)

PR #583, opened 2020-10-23
chore: convert test code to TypeScript

trusktr:
This is good for checking our own type definitions make sense. Some types were updated based on their usage in the tests.


package.json
```diff
@@ -31,7 +31,7 @@
 		"tsc": "tsc",
 		"tsc-watch": "tsc --watch",
 		"examples": "npx serve .",
-		"test": "npm run build && npm run test-unit && npm run test-lint",
+		"test": "npm run build && npm run test-lint && npm run test-unit",
```
trusktr, line 34: The output is more enjoyable when the last output on the screen is the green `OK: 420 assertions (31ms)` message of everything passing.

Issue #555, opened 2020-07-02
Dynamic_to does not work with array values


lguminski:
> Specify the revision number of the tween.js library where the bug occurred

Official release 17.4.0

> Specify your browser version and operating system (i.e. Chrome 23.0.1271.95, Windows 7)

Chrome 83.0.4103.97 (Official Build) (64-bit) Mac OS

> Describe the problem in detail. What happened? What did you expect to happen?

07_dynamic_to.html shows a situation when an object (fox) is tweened to a dynamic target (rabbit). It works, but this is a simple case when target is an absolute value. When the target is a dynamic array of 2 rabbits, then only the initial value of the array is taken into account.

> Provide a small test case

Please take a look at https://jsfiddle.net/lguminski/9ahnf8u6/ . In the example fox chases 2 rabbits. 

    var rabbits = {
      x: [rabbit1.x, rabbit2.x],
      y: [rabbit1.y, rabbit2.y]
    };

    new TWEEN.Tween(fox).to(rabbits, 3000)
      .interpolation(TWEEN.Interpolation.CatmullRom)
      .start()

and each rabbit dynamically updates its position

    new TWEEN.Tween(rabbit1).to({
      x: width - 50,
      y: height - 50
    }, 3000).onUpdate(function(object) {
      rabbits.x[0] = object.x;
      rabbits.y[0] = object.y;
    }).start();

    new TWEEN.Tween(rabbit2).to({
      x: width - 50,
      y: height - 50
    }, 3000).onUpdate(function(object) {
      rabbits.x[1] = object.x;
      rabbits.y[1] = object.y;
    }).start();

Unfortunately fox does not take the changes into account, and visits only initial positions of rabbits.

This issue seems to be similar to #545

trusktr:
Hello! Thanks for reporting this!

The issue in #545 happened in v18.6.0 and seems unrelated.

Did this work for you before?

trusktr:
If this never worked before, it might be a new feature.

lguminski:
Hello @trusktr , thank you for your feedback! 

I am pretty sure it worked before., and indeed, after checking, `17.3.0` works fine. Please check this demo: 

https://jsfiddle.net/lguminski/zsmtd5jq/

So `17.4.0` must have broken it.

And let me note that the user guide doesn't state that dynamic target and array values cannot be combined. So I am literally following the guide.

trusktr:
[...]

PR #585, opened 2020-10-26
Fix dynamic to with array values

trusktr:
fixes https://github.com/tweenjs/tween.js/issues/555 (in particular, see the plan at the bottom, and note that the fix for v17 has already been published in v17.6.0).

target: v20

@lguminski This does not include your example that modifies the interpolation array length. I think this is a good first step.

I'm planning to make some re-factors that would make it easier to implement that later, if we want it, as an additional feature.

This is a breaking change because now modifying an object outside of an animation is no longer dynamic by default. Previously, it was possible to dynamically modify the object passed to `to()` as in the "dynamic to" example where fox chases a rabbit, but it was not possible to dynamically modify interpolation arrays. Now, by default nothing is dynamic (for objects and interpolation arrays) and objects will not be modified, unless `dynamic` is set to `true`.

@mikebolt @lguminski @dalisoft What are your thoughts about how it currently works? Regardless if we make the default value of `dynamic` be `true` or `false`, either way is still a breaking change. Question is, which default value do you like more?

Currently default is `false`.

Note, in previous PRs, I converted `tests.js` to `tests.ts` (we can verify our types while we write tests, yay!), and I also previously added a test for `dynamic to` modifying an object (search for `dynamic` in `tests.ts`).

---

### additional idea

I am also thinking that a further improvement would be that instead of injecting starting values into interpolation arrays, we can simply read the starting values from `_valuesStart`, and hence we would never need to modify the objects passed into `to()` but still achieve the dynamic or non-dynamic behavior depending on the boolean value. With this approach, the only code that would ever modify an object passed into `to()` would be the end user's own code, never our Tween code.


trusktr:
I'm going to try the `additional idea` I mentioned above first. If it isn't too tricky, then it will:

- eliminate the need for a `dynamic` option (the dynamic-to feature will always be available)
- prevent Tween from ever writing to the user's object that was passed into `tween.to()`, except in the case that values are relative string values (e.g. `+20`).

We could additionally add an option like `hasRelative` that would be required in order to opt into the relative value feature (which would modify the user's object).

There's no way to allow both the relative-values and dynamic-to features while never writing to the end user's `to` object. At least, I don't yet see a simple way to do it. If user wants to use relative values, we can copy the object then apply the relative values, but that means we are using a copy so dynamic to doesn't work as the values are not applied to the copy. If they user wants dynamic values, and we don't copy the user's object, then relative values will be applied to the user's object and modify it.

lguminski:
I arrived late to the party. Sorry about that. I just [commented](https://github.com/tweenjs/tween.js/issues/555#issuecomment-822759895) under the original issue, and now I am seeing here that independently you arrived to similar conclusion. 

> I'm going to try the additional idea I mentioned above first. If it isn't too tricky, then it will:
> - eliminate the need for a dynamic option (the dynamic-to feature will always be available)
> - prevent Tween from ever writing to the user's object that was passed into tween.to(), except in the case that values are relative string values (e.g. +20).

First, I am happy to see it, because I did not like the `dynamic` option. Secondly I also agree that the user's object that was passed into tween.to() should not be altered by tween, as it violates ownership over memory. It is fine if tween manages its own buffers/caches, but user's original data must not be altered, as we don't know what one plans to do with the data later. Perhaps the data is used later after the tween finishes, and then the fact that tween changed it breaks code.

@trusktr  summing up, I believe that these are right choices.

trusktr:
This has been merged into `main` and released in [v20.0.0](https://github.com/tweenjs/tween.js/releases/tag/v20.0.0).

PR #589, opened 2020-11-02
Update yoyo function in user guide

JackParkes1:
The yoyo function currently has no code block example. This leaves it up to the user to guess whether calling the function sets yoyo to true or if a value needs to be passed in without referring to the code. Added lines to show that boolean values should be passed in to the function.


trusktr:
Awesome, thanks @JackParkes1 !

PR #591, opened 2020-11-08
expose duration with getDuration method

andrewkolos:
Since the `chain` method doesn't create a new animation object (it just starts the given tween whenever the current one finishes), I wanted to create my own mechanism to manage a sequence of tweens. However, there is no way to know a duration of a Tween after it is created, which is information I need to manage a sequence properly.

This PR adds a `getDuration()` method to `Tween` that gets the duration. It is analogous to the `getId` method.

### DIFF
--- a/dist/index.d.ts
+++ b/dist/index.d.ts
@@ -91,7 +91,7 @@ declare module "TWEEN" {
         static nextId(): number;
     }
 
-    const VERSION = "18.5.0";
+    const VERSION = "18.6.0";
 
     /**
      * Controlling groups of tweens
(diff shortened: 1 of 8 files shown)


trusktr:
This is a rather harmless change. If you'd want to get it in we still can. I can't re-open because the forked repo was deleted.

trusktr:
Also this is yet another example of why a more robust `Timeline`/`Keyframe`/`Tween`  API may be better; `Tween` does a lot by itself.

trusktr:
In fact I see you started something like that in `ts-tween`! Welcome any suggestions. :)

Issue #181, opened 2014-12-22
Getting tween state in the middle


khurramraheel:
I know its a silly question but is it possible to move the tween object my assigning some time value, like if I have a tween which has a duration of 10 seconds, Is there any way to start the tween and directly move to my desired time , for example what If I want to move my tween to move to 6 seconds or say 5 seconds and start from there?

sole:
Not right now with the current API, unless you did some sort of 'hack' like starting the tween at a t=0 and then immediately updating to t=5 and continuing from there. I haven't done it myself.
Sorry about the delay, I missed the notifications.

humodz:
You can pass an argument to `tween.start()`:
```js
// Sets the start time to 6 secs in the past
myTween.stop().start(TWEEN.now() - 6000)
// or
myTween.stop().start(String(-6000));
```

Full example:
https://jsbin.com/mayafehoye/11/edit?html,js,output

PR #597, opened 2021-01-25
Add new method Tween.startFromCurrentValues

chris-bingham:
WIP as I havent added an example yet, I will if you are happy with the addition to the API.

## What? 
New method: `Tween.startFromCurrentValues(time?: number)`
* Starts the tween but using the current values of the tweened properties, rather than the values the first time start was called
* This calls start but passes `true` to an added second param: `overrideStartingValues` which is passed through to `_setupProperties`

## Why?
We are using `tween.js` in a game engine where expected behaviour of animations is for them to start from the values held by the sprites when the animation starts. Animations can be run multiple times, and triggered by vast number of events. The animated sprites could have completely different values on subsequent starts than they did the first time start was called.
I believe this was formerly the behaviour of start?

### DIFF
--- a/examples/18_start_from_current_values.html
+++ b/examples/18_start_from_current_values.html
@@ -1,0 +1,110 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<title>Tween.js / hello world!</title>
+		<meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+		<link href="css/style.css" media="screen" rel="stylesheet" type="text/css" />
+	</head>
+	<body>
+		<div id="info">
+			<h1><a href="http://github.com/tweenjs/tween.js">tween.js</a></h1>
+			<h2>18 _ start from current values</h2>
+			<p>Example for illustrating the startFromCurrentValues() functionality.</p>
+			<button id="start-tween">
+				Start tween
+			</button>
+			<button id="up">
+				↑
+			</button>
+			<button id="down">
+				↓
+			</button>
+			<button id="left">
+				←
+			</button>
+			<button id="right">
+				→
+			</button>
+		</div>
+		<div
+			id="target"
+			style="
+				position: absolute;
+				top: 200px;
+				left: 400px;
+				width: 100px;
+				height: 100px;
+				background: #a0dde9;
+				padding: 1em;
+			"
+		></div>
+
+		<script src="../dist/tween.umd.js"></script>
+		<script src="js/RequestAnimationFrame.js"></script>
+		<script>
+			var position
+			var target
+			var tween, tweenBack
+
+			init()
+			animate()
+
+			function init() {
+				position = {x: 400, y: 200}
+				target = document.getElementById('target')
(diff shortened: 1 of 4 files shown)


trusktr:
This could be interesting. Let's see how it differs from the current ways to achieve the same.

trusktr:
One way to do it is to remember the last time value, and start there: https://github.com/tweenjs/tween.js/issues/181#issuecomment-778783709

The time value itself can also be animated with the `.update` methods.

But I agree that it wouldn't be as ergonomic. We want to simply set the new destination values, start the tween, and pass it current time values.

chris-bingham:
Thanks for looking into this @trusktr. Remembering the time value is an interesting technique, but it wouldn't actually solve our problem. The animated properties may be changed between the tweens by factors other than Tween.js. 

For example take a dressup game in which the user drags clothes onto a different body parts:
1. The user drags a t-shirt and drops it so that the sleeve is just intersecting the torso
2. Since the t-shirt sprite intersects its intended target we class this as a successful placement so we use Tween.js to animate the t-shirt into the perfect position on the torso
3. The user then decides to drag the t-shirt off the torso but quickly changes their mind and puts it back, but this time they drop it intersecting the opposite side of the torso
4. We now want the tween to start from the current values rather than the starting values it held the first time the t-shirt was placed

S-unya (approved):
I wonder if it makes sense to try to install from the tween.js PR? It also worth noting that PR is currently a WIP... should that be updated as nothing has happened with it for a while

trusktr:
If existing tests pass and an example is added, let's merge this. I think it will be useful, at least for now (I'm thinking something like a Timeline and Keyframe API would be better longer term, with Tween being literally just a tween, whereas at the moment it has many things in it that basically make it Timeline, Keyframe, and Tween all in one).

Can you make an example @chris-bingham ?

chris-bingham:
Cool, yes will make an example either this week or next 👍

trusktr:
Updated with latest from master.

chris-bingham:
Hi @trusktr, really sorry for the delay on this!
 I have added an example. I felt the easiest way to demo this was to use a set of buttons, I've styled these in a similar fashion to the rest of the examples.

S-unya:
Hi @trusktr , is this PR still pertinent? As far as I can tell, the requirements have been met, but @chris-bingham is on sabatical, so I am happy to continue this work if you feel it is still useful.

trusktr:
Sorry for the delay. Merged! Thanks for this very useful change!

Are either of you @S-unya @chris-bingham interested in joining the org to help to maintain and expand Tween.js?

S-unya:
@trusktr thanks very much for the offer. In principle I am keen; I will need to discuss it with the team and get back to you if that is ok?

trusktr:
Yeah of course! In the meantime v20 is out!

S-unya:
@trusktr, just to keep you informed, development time being provided to help tween.js is still being discussed... everybody seems happy with the idea, but there are some dilligence criteria that need discussing above my head.

S-unya:
@trusktr I have go ahead for members of our team to work on tween.js if you still want us to help.

PR #598, opened 2021-02-18
fix: end()ed tweens would leak and remain as zombies (unkillable with stop()), and would rise again from the dead if before the delayTime

ProPuke:
Tweens that reach their end or have `end()` called on them are not removed from their group, but do have their `_isPlaying` set to false, meaning they cannot then be removed with `stop()`, forever stuck between life and death with their `update()` still always called.

This means that trying to jump a tween to the end early and then stop it (`tween.end().stop()`) would instead leave it as a zombied instance.

As a bonus effect, calling `tween.end().stop()` on a tween with a `delay()`, which has not yet passed would result in the tween autostarting later on and raising from the dead.

### DIFF
--- a/src/Tween.ts
+++ b/src/Tween.ts
@@ -418,6 +418,9 @@ export class Tween<T extends UnknownProps> {
 					this._chainedTweens[i].start(this._startTime + this._duration)
 				}
 
+				// eslint-disable-next-line
+				this._group && this._group.remove(this as any)
+
 				this._isPlaying = false
 
 				return false


ProPuke:
My bad, this problem seems to only manifest when a delay() is present and has not yet autostarted, and the order of calling is `.end().stop()`. I guess this falls under correct behaviour then, since if it has a delayed start and you tell it to end, it will jump to the end of its life, but cos it isn't technically active yet it can't be stopped, then when the delay passes it comes to life.

I should have been doing `.stop().end()` instead for the "correct" behaviour. This tastes a bit odd and initially counter-intuitive, but makes sense now on examination :S

trusktr:
Yet another issue that makes me think we need a better `Timeline`/`Keyframe`/`Tween` abstraction. At the moment `Tween` tries to do everything.

Issue #378, opened 2017-09-17
Animation on repeat speeds up over time.


trusktr:
Here's @jonobr1's original, forked to use array values:

https://codepen.io/trusktr/pen/04518ee24a8d1d7822e584c6a513d378?editors=1010

Watch it for a while. Over time, the animation speeds up, which shouldn't happen.

jonobr1:
Do you know what the bug I'm leveraging is?

trusktr:
@jonobr1 Oops, I meant to post a link to a fork of yours, where the only difference is that I'm passing an array of values to `o`'s `to` `rotation`, on line 10:

http://jsfiddle.net/18ymqpdw/0/

If you watch it for a while, you'll notice that the animation seems to speed up after a few repeats.

I updated the original post to use the forked link. What bug were you leveraging in [yours](
http://jsfiddle.net/jonobr1/72bytkhm/)?

jonobr1:
I'm not too familiar with arrays in Tween.js..., but if that's the difference then I don't think I'm leveraging any bugs in the fiddle I made. Also, this might be an inaccurate use of the API and may not be a bug.

dalisoft:
@trusktr See [fiddle](http://jsfiddle.net/s8bjk6r0/1/) using es6-tween, but it has bug also

trusktr:
Interesting, at least that behavior is consistently repeating, rather than changing each time it repeats. Let's see what's going on here..

dalisoft:
@trusktr It is about tween.js or es6-tween. I am added this to es6-tween issue list and i will fix as soon as possible

dalisoft:
@trusktr Is in this [fiddle](http://jsfiddle.net/18ymqpdw/3/) issue was fixed?

trusktr:
Looks like the examples broke. I update the OP with a working example on codepen. I'd like to start fixing stuff and moving forward!

trusktr:
@dalisoft Mind updating your example?

Also, sidenote, if you don't mind making some PRs to bring your ideas into the main repo, that'd be great!

dalisoft:
@trusktr I [updated](http://jsfiddle.net/64vdmw7u/), but seems does not work, some incompatibility with two.js

I deprecated es6-tween due of less time to spent, i think these some things can be applied and source code can be got from es6-tween

- TypeScript (currently es6-tween is ES6, but before have been written TS)
- DOM Plugins
- Infinite Deep-Level tweening without loss of performance
[...]

PR #602, opened 2021-04-14
Multiple `.start()` and `.to(prop:[])`

MasatoMakino:
## Related Issues

- Animation on repeat speeds up over time. #378

This PR fixes the Issue of tween acceleration on multiple .start().

If an object with an array is passed to `.to()`, `_valuesEnd` will be irreversibly changed when `.start()` is executed multiple times.
As a result, each time you run `.start()`, animation will be faster.

## Details of the changes

1. Added test cases.
1. Added `_hasSetupProperty` flag to check if `_setupProperties()` has been executed.
1. If you reconfigure `.to()`, startValue will be changed to the current value of the tween target object. Check `_valuesStart` and keep it if it is already initialized.

https://github.com/tweenjs/tween.js/blob/ae24c58c8b57b2079151ae2c3c63f2bb501c8b2b/src/Tween.ts#L169-L172

I used this code as a reference for change point 3.

## Related pull requests

- Fix dynamic to with array values #585

I think it might be difficult to merge this PR as-is, because of conflicts with features such as `dynamic to` and `_propertiesAreSetUp`. In that case, I hope the added test cases will contribute to everyone.

### DIFF
--- a/src/Tween.ts
+++ b/src/Tween.ts
@@ -44,6 +44,7 @@ export class Tween<T extends UnknownProps> {
 	private _onStopCallback?: (object: T) => void
 	private _id = Sequence.nextId()
 	private _isChainStopped = false
+	private _hasSetupProperty = false
 
 	constructor(private _object: T, private _group: Group | false = mainGroup) {}
 
@@ -65,6 +66,7 @@ export class Tween<T extends UnknownProps> {
 		// currently no opt-out).
 		// for (const prop in properties) this._valuesEnd[prop] = properties[prop]
 		this._valuesEnd = Object.create(properties)
+		this._hasSetupProperty = false
 
 		if (duration !== undefined) {
 			this._duration = duration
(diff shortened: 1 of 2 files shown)


trusktr:
@MasatoMakino Thanks for this! Would you mind making an example for this?

trusktr:
The more I see changes and fixes like these, the more I start to want to start working on a Timeline/Keyframe/Tween abstraction. It seems Tween is trying to do too much, all in one.

MasatoMakino:
> @MasatoMakino Thanks for this! Would you mind making an example for this?

@trusktr Thanks for your reply.

I have made examples of multiple start with CodePen.
[tween.js 18.6.4 : multiple start and array](https://codepen.io/masatomakino/pen/bGgmGYa)
[tween.js 18.6.4 + PR #602 : multiple start and array](https://codepen.io/masatomakino/pen/OJWoKrq)

I also considered adding a new page in `./examples`. 
However, it is hard to tell what is a difference between examples of multiple start and [example of repeat](http://tweenjs.github.io/tween.js/examples/08_repeat.html), and there is a risk of confusing users. I did not add examples of multiple start to this PR.

trusktr:
Closes #378

trusktr:
Looks like this got fixed already!

https://codepen.io/trusktr/pen/XWGpKyX/10faa7d954bf016a3f0261ca626db17e?editors=1010

There's a similar issue with `startWithCurrentValues()`, but the problem is gone when using `start()`. See this issue:

- https://github.com/tweenjs/tween.js/issues/378

Issue #116, opened 2014-05-03
Tween jumping to finish position


KeeleyMoore:
Hi, thanks for this great lib its really useful! 

But i've fallen into some trouble that i cant figure out. Within my three.js scene i have camera and avatar objects in a container object that is moved with my controller. On collision with another object in my scene the camera is meant to tween back and up slightly but instead of any animation it just jumps to the finished position.

This only happens while the avatar is being moved, if i initiate the tween straight away when the avatar isn't being moved the animation runs fine. 

Is anybody aware of what might be causing this? and a possible fix? I'm truly stumped

This is the tween setup code i'm using incase its just a simple setup issue i've missed:

```
    var camPosition = { x : 0, y : 5, z : 50 };
    var camTarget1 = { x : 0, y : 10, z : 75};

    tweenCamToOrig = new TWEEN.Tween(camTarget1)
    .to(camPosition, 2000)
    .easing( TWEEN.Easing.Back.InOut )
    .onUpdate(function(){
        camera.position.x = camPosition.x;
        camera.position.y = camPosition.y;
        camera.position.z = camPosition.z;
    })
    .onComplete(function(){
    });
```

rbary:
Thanks a lot  KeeleyMoore this post helped me !!

sole:
Perhaps we need to make an example / mention this in the docs!
@rbary what was your issue exactly and how did this issue or post link help you? I'd like to clarify whatever is it that is not clear! thanks!

rbary:
Hi my issue was perspective camera (three) animation;

[https://github.com/NephtysOrg/projet_sia/blob/other_dev/js/app/game.js]()

````javascript
Game.prototype.cameraTranslate = function(){
    var actualXpos =this.current_camera.position.x;
    var actualYpos =this.current_camera.position.y;
    var actualZpos =this.current_camera.position.z;
    console.log(this.current_camera.position);
    var differenceX = Math.abs(this.current_camera.position.x - actualXpos);
    actualXpos = this.current_camera.position.x ;
    var differenceY = Math.abs(this.current_camera.position.y - actualYpos);
    actualYpos = this.current_camera.position.y;
    var differenceZ = Math.abs(this.current_camera.position.z - actualZpos);
    actualZpos = this.current_camera.position.z;

    this.current_camera.position.x += differenceX;
    this.current_camera.position.y += differenceY;
    this.current_camera.position.z += differenceZ;    
};
Game.prototype.cameraTransition = function (position,look) {
[...]

PR #604, opened 2021-04-22
Add pow easing function

MasatoMakino:
## Related Issues

- tween jumps (jitters) to the final value at the end of animation #116

This PR is for adding an easing function that has no jitter and a sharp curve like `TWEEN.Easing.Exponential`.

## Details of the changes

1.  Added `TWEEN.Easing.generatePow()` function.
2. Added test cases.
3. Added example page.

`TWEEN.Easing.generatePow()` function the changes sharpness of a curve depending on `power` argument. This approach is the same as the [expo easing function](https://github.com/Popmotion/popmotion/blob/a1903c808757b31d26d876dc2d8d3dea2d131c12/packages/popmotion/src/easing/utils.ts#L14) in [popmotion](https://github.com/Popmotion/popmotion).

[GSAP](https://greensock.com/docs/v3/Eases) calls the similar easing function Power1 ~ 4. Also, [CreateJS/TweenJS](https://github.com/CreateJS/TweenJS) internally calls this easing function [getPowIn, Out, InOut](https://github.com/CreateJS/TweenJS/blob/6df2b9ce7e68d3d824cc490b67cab2bc3204810a/src/tweenjs/Ease.js#L105). Following these precedents, I named this function generatePow.

### DIFF
--- a/src/Easing.ts
+++ b/src/Easing.ts
@@ -195,6 +195,30 @@ const Easing = {
 			return Easing.Bounce.Out(amount * 2 - 1) * 0.5 + 0.5
 		},
 	},
+	generatePow: function (
+		power = 4,
+	): {
+		In(amount: number): number
+		Out(amount: number): number
+		InOut(amount: number): number
+	} {
+		power = power < 1.0 ? 1.0 : power
+		power = power > 10000 ? 10000 : power
+		return {
+			In: function (amount: number): number {
+				return amount ** power
+			},
+			Out: function (amount: number): number {
+				return 1 - (1 - amount) ** power
+			},
+			InOut: function (amount: number): number {
+				if (amount < 0.5) {
+					return (amount * 2) ** power / 2
+				}
+				return (1 - (2 - amount * 2) ** power) / 2 + 0.5
+			},
+		}
+	},
 }
 
 export default Easing
--- a/src/Easing.ts
+++ b/src/Easing.ts
@@ -202,7 +202,7 @@ const Easing = {
 		Out(amount: number): number
 		InOut(amount: number): number
 	} {
-		power = power < 1.0 ? 1.0 : power
+		power = power < Number.EPSILON ? Number.EPSILON : power
 		power = power > 10000 ? 10000 : power
 		return {
 			In: function (amount: number): number {
--- a/examples/17_generate_pow.html
+++ b/examples/17_generate_pow.html
@@ -1,0 +1,71 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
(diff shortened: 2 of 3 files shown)


trusktr:
Awesome!

src/Easing.ts
```diff
@@ -195,6 +195,30 @@ const Easing = {
...
 		},
 	},
+	generatePow: function (
+		power = 4,
+	): {
+		In(amount: number): number
+		Out(amount: number): number
+		InOut(amount: number): number
+	} {
+		power = power < 1.0 ? 1.0 : power
```
trusktr, line 205: Just curious, why not allow numbers less than 1?

trusktr: I'm imagining that with a number less than 1, for example `generatePow(0.3).InOut`, the user can animate as if some fast-moving object comes into the view, then the "film speed" is slowed down and the object is in slow motion, then the "film speed" is sped back up to normal speed and the moving object takes off out of the view at full speed.

With a more exaggerated number like `0.1`, it could be as if a "UFO" was caught on camera: it comes into view and slows down very quickly, then suddenly accelerates out of view.

MasatoMakino: @trusktr 
Thanks for your advice.

I had misunderstood that easing breaks when power is below 1. 
https://www.desmos.com/calculator/in8brckvcq
After checking, I found a beautiful curve in range of 0.0 < power < 1.0.
Pow easing curve breaks down when power is below zero.

I will modify conditions to 0.0 < power.

trusktr: nice!

trusktr:
Hello @MasatoMakino! I can't find a value for `generatePow()` that behaves close enough to `Exponential`, I can always tell the difference visually. There's something pleasing about Exponential (except the jump at the end).

It would be awesome to get the same effect as our current Exponential, but without the jump (start 0 end 0). Here's a video showing `generatePow(5.8)` vs `Exponential`:


https://user-images.githubusercontent.com/297678/127800717-c42d8e1d-2cd3-48ca-8267-b81a52e05b5d.mp4

MasatoMakino:
Hello @trusktr.

First of all, please note that `generatePow()` function cannot return exactly the same result as `Exponential`. A sample easing curve can be found in https://github.com/tweenjs/tween.js/issues/116#issuecomment-820943726 

https://www.desmos.com/calculator/xbgt04gh4v

A parameter of 5.8 is the value at which the intersection of the two curves is at the center.
For a softer landing, setting the parameter to about 5.3 ~ 5.4 looks good.
If you are looking for a sharp rise, 6.4 ~ seems to be a good setting.

I hope this information will help you.

PR #607, opened 2021-04-28
add setProgress to Tween

kanbang:
add Tween.setProgress to set current progress, then call update callback


trusktr:
@kanbang Hello. Thanks for making this idea. A couple things:

1. Is it worth adding? We can currently update the tween by manually calling `update`
2. The new code has a duplicate of the following for the most part:
   https://github.com/tweenjs/tween.js/blob/d8b4f09fb3e2f56676d262ba139391eb3b486b10/src/Tween.ts#L361-L371
   If we were to get this in, we would need to consolidate the duplication so that it is easy to manage and won't break on accident (for example if someone changes one part but not the other).

optimus007:
this would be handy

- set progress to 0.5 then call .start() it starts from this point and the easing continues from 0.5 and with half the duration 
- this can be used to scrub a tween
- this will be useful when there's a very long tween and you just to playback the last quarter or from a certain section

trusktr:
I can see how it might be useful, but also trying to avoid complexity. The behavior of going backwards in time is not very well defined (f.e. other features like yoyo, repeat, etc, might break in unexpected ways).

For now, animating the time value backwards for a single tween should work. This is good enough for now. Just map a value between 0 and 1 to the time range.

When we eventually implement [`Timeline`](https://github.com/tweenjs/tween.js/issues/647), we'll have a simple setup without everytying being tightly coupled in a single class, and scrubbing will be easier and cleaner. Closing in favor of going that direction (and for now just map your time values on a single Tween but don't expect all features to work with negative time progression).

trusktr:
@kanbang thanks for trying this feature!

Issue #507, opened 2019-11-06
Help add new easing functions


mikebolt:
We could always use more easing functions.

An easing function is single-variable function that maps values in the range [0, 1] to a numeric value. Generally, we require that `f(0) = 0` and `f(1) = 1`, but otherwise there are no constraints.

If you want to help out feel free to submit a PR, and don't worry about code style. If you need help I can provide a template for your PR.

Tests aren't required for easing functions, but if you want to you can add the new easing function to the graphs example.

trusktr:
Maybe something meat would be an easing function primitives for creating easing functions, like what popmotion has.

dalisoft:
@mikebolt What about https://www.npmjs.com/package/owneasing? Maybe helps user create easing by curving which can be created by almost any SVG editor

dalisoft:
Or this also may be useful https://www.npmjs.com/package/es6tween-physics

trusktr:
Nice! Let's evaluate how we may incorporate those.

trusktr:
For reference, here's the functions in popmotion for creating new curves: https://popmotion.io/api/easing.

trusktr:
Cleaning issues. Closing as there is not a concrete actionable here. Feel free to open new issues with specific easing curves that may be desired.

PR #608, opened 2021-04-30
Add unit test for easing function

MasatoMakino:
## Related Issues

- Help add new easing functions #507
- tween jumps (jitters) to the final value at the end of animation #116

This PR is for adding unit tests to TWEEN.Easing functions.

### 'Test TWEEN.Easing should starts at 0.0, ends at 1.0. TWEEN.Easing.InOut() should be 0.5 at midpoint'

Based on #507, this test checks that all easing functions satisfy conditions `f(0) = 0.0`, `f(1.0) = 1.0`. This test also checks that `Easing.InOut(0.5) = 0.5`.

### 'Test TWEEN.Easing should pass a specific value'

This test verifies that a result of `f(Math.LOG10E)` is not broken. The purpose of this test is to check if results of the function changes depending on the version of node.js.

`toBeCloseTo` are based on [jest's functions](https://jestjs.io/docs/expect#tobeclosetonumber-numdigits).

### fix Sinusoidal easing

https://github.com/tweenjs/tween.js/blob/d8b4f09fb3e2f56676d262ba139391eb3b486b10/src/Easing.ts#L73

Because `1 - Math.cos(1.0 * Math.PI / 2) = 0.99999999999` , it does not snap to 1.0.
By replacing Math.cos with Math.sin, this function snap to 1.0.

### fix Back easing

Added truncation of 1.0 and 0.0.

### DIFF
--- a/src/Easing.ts
+++ b/src/Easing.ts
@@ -70,13 +70,13 @@ const Easing = {
 	},
 	Sinusoidal: {
 		In: function (amount: number): number {
-			return 1 - Math.cos((amount * Math.PI) / 2)
+			return 1 - Math.sin(((1.0 - amount) * Math.PI) / 2)
 		},
 		Out: function (amount: number): number {
 			return Math.sin((amount * Math.PI) / 2)
 		},
 		InOut: function (amount: number): number {
-			return 0.5 * (1 - Math.cos(Math.PI * amount))
+			return 0.5 * (1 - Math.sin(Math.PI * (0.5 - amount)))
 		},
 	},
 	Exponential: {
@@ -159,11 +159,11 @@ const Easing = {
 	Back: {
 		In: function (amount: number): number {
 			const s = 1.70158
-			return amount * amount * ((s + 1) * amount - s)
+			return amount === 1 ? 1 : amount * amount * ((s + 1) * amount - s)
 		},
 		Out: function (amount: number): number {
 			const s = 1.70158
-			return --amount * amount * ((s + 1) * amount + s) + 1
+			return amount === 0 ? 0 : --amount * amount * ((s + 1) * amount + s) + 1
 		},
 		InOut: function (amount: number): number {
 			const s = 1.70158 * 1.525
--- a/src/tests.ts
+++ b/src/tests.ts
@@ -630,6 +630,34 @@ export const tests = {
 		test.done()
 	},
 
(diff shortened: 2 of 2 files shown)


src/Easing.ts
```diff
@@ -159,11 +159,11 @@ const Easing = {
 	Back: {
 		In: function (amount: number): number {
 			const s = 1.70158
-			return amount * amount * ((s + 1) * amount - s)
+			return amount === 1 ? 1 : amount * amount * ((s + 1) * amount - s)
```
trusktr, line 162: Does this fix a jump?

trusktr:
Hey, nice stuff! What do you mean by "The purpose of this test is to check if results of the function changes depending on the version of node.js."?

src/tests.ts
```diff
@@ -1954,3 +2054,20 @@ type Test = {
 	expect(n: number): void
 	done(): void
 }
+
+type EasingFunctionGroup = {
+	In(amount: number): number
+	Out(amount: number): number
+	InOut(amount: number): number
+}
```
trusktr, line 2062: Cool. I suppose this can be moved to the definition of the function groups in https://github.com/tweenjs/tween.js/blob/master/src/Easing.ts, then imported here.

src/Easing.ts
MasatoMakino: Yes.
`Back.In(1.0) = 0.9999999999999998`
This jump is not visually problematic, but it is not enough to pass the test.
As the jump is very small, I did not modify the process and added a truncation.

src/tests.ts
MasatoMakino: During the writing of this test, I considered how to add the definition to `Easing.ts`.

In order to add the EasingFunctionGroup type definition, we had to modify the export of index.ts.
I thought it would be excessive to modify index.ts to add tests, so I added a type definition to test.ts.
In the future, I also think that adding a type definition to Easing.ts is the best way to go.

MasatoMakino:
> Hey, nice stuff! What do you mean by "The purpose of this test is to check if results of the function changes depending on the version of node.js."?

If node.js is updated and the decimal point handling or Math class handling changes, this test will detect the change.

CI of tween.js supports multiple versions of node.js. Therefore, it is easy to determine which version the change occurred in.

A possibility of such a disruptive change in node.js is very low. However, tracking with CI does not waste developers' time. I think it's a good deal to add a few dozen lines of test code to save our collaborators' valuable time.

Issue #456, opened 2019-05-27
Add a test for onRepeat


mikebolt:

trusktr:
closed in #609. Thanks @MasatoMakino 🙏 !

Issue #290, opened 2016-09-13
Implement onBeforeRepeat


shelbeniskb:
is it possible to offer onRepeat? I need to change some state before the repeat cycle start, so I onRepeat is necessary, I also find TweenMax offer this function.

sole:
Could you define a bit better how 'onRepeat' would work? Would it be executed before each repeat cycle? Would it maybe be better called `onBeforeRepeat`?

dalisoft:
I like `onBeforeRepeat` and `onBeforeReverse`. Very good naming.

shelbeniskb:
@sole `onBeforeRepeat` will be an chance for developer to change some state or value, for example,
move a picture from A to B,  we can change the picture before the next repeat happens.

sole:
Ok, makes sense. I will update the title of the issue to reflect this :)

dalisoft:
@shelbeniskb Maybe you looking for [that](https://github.com/tweenjs/es6-tween/blob/master/dist/Tween.js#L1055)? It implements `repeat and reverse` events, you can define via `tween#on('reverse', (object) => { console.log(object); })`

trusktr:
An `onRepeat` feature has since been added. If that's not enough, definitely feel free to reply and we can re-open of make a new issue. :)

PR #609, opened 2021-05-02
Add : tests for onRepeat callback

MasatoMakino:
## Related Issues

- Add a test for onRepeat #456 
- Implement onBeforeRepeat #290

This PR is to add tests for the Tween.onRepeat callback.

## Details of the changes

I add the following tests. This test follows current onRepeat implementation.

1. TWEEN.Tween.onRepeat should not be called if repeat = 0 or default
2. TWEEN.Tween.onRepeat should be called once if repeat = 1
3. TWEEN.Tween.onRepeat should be called every time if repeat = Infinity
4. TWEEN.Tween.onRepeat should not be called if Tween.pause() or Tween.stop(), and should be called after Tween.resume() or restart
5. If Tween.delay is set, TWEEN.Tween.onRepeat should be called when repeat section finished
6. If Tween.repeatDelay is set, TWEEN.Tween.onRepeat should be called when repeat section finished

## Confirmation

I found that onRepeat behaves differently from [GSAP](https://greensock.com/gsap/). This difference appears when we set delay or repeatDelay.

GSAP : onRepeat is called just before the tween finishes, waits for a delay, and then plays the next repeat.
tween.js : onRepeat is called just after the tween finishes, before waiting for the delay.

[onRepeat by gsap](https://codepen.io/masatomakino/pen/zYNVVXr)
[onRepeat by tween.js](https://codepen.io/masatomakino/pen/vYgqovv)

I wrote a codepen on this confirmation. Repeat count on the screen is increased by onRepeat. You can see that the timing at which the repeat count changes is different for each sample.

Is this difference as per specifications? Or is it a bug?
I will keep this PR in draft form until this point is confirmed.


trusktr:
Updated with `main`

trusktr:
Closes https://github.com/tweenjs/tween.js/issues/456

trusktr:
merged finally!

Issue #610, opened 2021-05-03
Properties of Easing object can reassignment


MasatoMakino:
## Description

All properties of the `TWEEN.Easing` object allow reassignment.

## Expected Behavior

Easing functions and groups in `TWEEN.Easing` do not allow reassignment.

## Actual Behavior

Easing functions and groups in `TWEEN.Easing` allow reassignment.

## Steps to Reproduce

Open [code pen](https://codepen.io/masatomakino/pen/NWdQjqx)

This is a problem section.

```js
TWEEN.Easing.Quadratic = TWEEN.Easing.Linear;
TWEEN.Easing.Cubic.In = TWEEN.Easing.Bounce.Out;
```

This code can be run without causing any errors.

## Impact

In any scope, `Easing` can be reassignment. If someone accidentally reassignment easing functions, it will be difficult to identify the problem.

## Proposal

Ideally, if we try to reassign to TWEEN.Easing, we will get an error.
Even if no error is thrown, this issue will be resolved if the reassignment is ignored.

- Remove `Easing` global object
- Remove default export in Easing.ts
- Change each EasingFunctionGroup (e.g. Quadratic, Cubic) to Class.
- Change each EasingFunction to static functions
- Change Easing import in index.ts to `import * as Easing from ". /Easing.ts"`

With this change, I think it will possible to ignore reassignments without losing compatibility.

## Environment

macOS 10.15.7
Google Chrome 90.0.4430.93
tween.js 18.6.4

trusktr:
What if we apply Object.freeze?

MasatoMakino:
I compared the two approaches. `Object.freeze` and `static functions`

## Points in common between the two approaches

### Both of these approaches use Object.freeze. 

If we remove a default export of Easing.ts, this code transpiled by tsc will look like this. You can see that `Easing` is frozen.

```js
var Linear = /** @class */ (function () {
    function Linear() {
    }
    Linear.None = function (amount) {
        return amount;
    };
    return Linear;
}());
// export default Easing

var Easing = /*#__PURE__*/Object.freeze({
    __proto__: null,
    Linear: Linear
});
```

Therefore, when we try to change a property, the behavior is the same.

### Compatibility can be confirmed by testing

Tests added in #608 will confirm that Easing keeps a compatibility after modify.

## Object.freeze

### Advantage

- No change to existing code

### Disadvantage

[...]

PR #617, opened 2021-05-27
fix : freeze easing objects

MasatoMakino:
## Related Issues

Properties of Easing object can reassignment #610
Add unit test for easing function #608 [comment](https://github.com/tweenjs/tween.js/pull/608#discussion_r624423020)

This PR is to block any reassignment of Easing objects.

## Details of the changes

- Freeze Easing object.
- Add unit test.


trusktr:
mentioned in the issue, lets keep Object.freeze, but restore the same export format so it isn't breaking,

PR #619, opened 2021-08-02
`Easing.generateExponential()`

### LINKED ISSUE #116: Tween jumping to finish position

trusktr:
Given a value of 300, this is very close Easing.Exponential, but it will hit 0 when amount is 0, therefore being an alternative to Exponential to avoid the jump issue mentioned in https://github.com/tweenjs/tween.js/issues/116.

I haven't tried to simplify the formulas yet. Also, could there be a better formula?

The video shows how similar the two are:

https://user-images.githubusercontent.com/297678/127811339-80ab8a98-4cf5-4ea2-96f4-ba4cfec76d02.mp4

- [ ] Add tests
- [ ] Add docs

@MasatoMakino Wdyt?

### DIFF
--- a/examples/00_hello_world.html
+++ b/examples/00_hello_world.html
@@ -4,56 +4,63 @@
 		<title>Tween.js / hello world!</title>
 		<meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
 		<link href="css/style.css" media="screen" rel="stylesheet" type="text/css" />
+		<style>
+			.threedee {
+				position: absolute;
+				transform-style: preserve-3d;
+				transform: translate3d(0, 0, 0.1px);
+			}
+			.box {
+				width: 100px;
+				height: 100px;
+				padding: 1em;
+				box-sizing: border-box;
+			}
+		</style>
 	</head>
-	<body>
+	<body style="transform-style: preserve-3d; perspective: 800px;">
 		<div id="info">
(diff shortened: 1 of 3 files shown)


MasatoMakino:
Great idea! I couldn't come up with that function.
If you plan to add testing as well, I have nothing to comment on.

PR #622, opened 2021-10-02
`onStartEvery`, as `onStart` but called on every repeat

tobyspark:
Draft PR –
If this sounds useful, I can work this up per guidelines. If this isn’t wanted, the commit has already served it’s purpose: it's in use and critical to my current `pixi.js` project.

Feature –
This is similar but not equivalent to `onRepeat`. The behaviour is as `onStart`, but it will trigger on every repeat. Unlike `onRepeat`, this means it will trigger at the start of the animation cycle, not the end of the previous. This is a critical difference if delays are in use. It also may give more expected results depending where in the draw cycle the TWEEN.update is called.

### DIFF
--- a/src/Tween.ts
+++ b/src/Tween.ts
@@ -38,6 +38,8 @@ export class Tween<T extends UnknownProps> {
 	private _chainedTweens: Array<Tween<any>> = []
 	private _onStartCallback?: (object: T) => void
 	private _onStartCallbackFired = false
+	private _onStartEveryCallback?: (object: T) => void
+	private _onStartEveryCallbackFired = false
 	private _onUpdateCallback?: (object: T, elapsed: number) => void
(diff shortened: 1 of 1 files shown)


trusktr:
So the difference is that it fires one extra time (at the very beginning with onStart) plus every other start (at the same time as onRepeat)? I think it makes to have it.

We would need a unit test for it too.

tobyspark:
The critical point is, that *isn’t* the same time as `onRepeat`. I’ll write up the docs, which should make that clear. And add a test!

tobyspark:
There you go, documentation and tests.

tobyspark:
I just reworked the when-what-is-called part of the documentation. It takes a lot more vertical space, but shows the differences precisely and with a code sample people can further experiment with.

PR #644, opened 2023-04-21
Improved yoyo documentation

iangilman:
It wasn't clear from the documentation that I had to pass anything into the `yoyo` function. I had to go look at the yoyo example page to figure that out. Hopefully this change will make it clear to the next person.

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -191,7 +191,11 @@ Check the [Repeat](../examples/08_repeat.html) example.
 
 ### `yoyo`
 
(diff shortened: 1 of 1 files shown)


trusktr:
Thanks! Good catch @iangilman!

trusktr:
Hmm, oh! Looks like the change is already here:

https://github.com/tweenjs/tween.js/blob/main/docs/user_guide.md#yoyo

We need to sync github pages, rather than update github pages directly.

Tracking issue: https://github.com/tweenjs/tween.js/issues/645

iangilman:
Funny! Glad I could help highlight the sync issue, though that was unintended 😄

Thank you for cherry picking my commit so I could get credit. ❤️

PR #646, opened 2023-04-23
docs(zh-CN): Update zh-CN docs & Add language options in the document

mk965:
I found that the Chinese and English documents are not consistent. I re-translated the Chinese document and added language options in the document.


trusktr:
Hey, thanks for taking this on! Looks like there is a conflict somehow.

trusktr:
@mk965 decide you don't want submit the change?

mk965:
> @mk965 decide you don't want submit the change?

Sorry for my misoperation. I will compare the v20 version of the English document and submit the Chinese document translation in a while

Issue #649, opened 2023-04-28
CJS entry point can't be loaded in Node.js


inukshuk:
I might be doing something wrong, but I believe there is an issue with the current dual package setup. Type is set to `module` but the CJS entry point `main` points to "dist/tween.cjs.js". Despite the `.cjs` in there the Node.js loader still sees the final `.js` extension. Based on the nearest `package.json` it will therefore default to using the ESM loader for this file.

As a result, you can't currently `require('@tweenjs/tween.js')` in Node. Unless I'm missing something [the `.js` extension should be dropped](https://github.com/tweenjs/tween.js/blob/main/rollup.config.js#L22).

trusktr:
Thanks @inukshuk and @christjt for pointing these issues out!

inukshuk:
Unfortunately the linked PR above doesn't address this issue. The CJS entry point needs to have a `.cjs` extension or the Node.js module loader will refuse to open it. The current `.cjs.js` one doesn't work.

inukshuk:
I submitted a PR for this above. It just renames the CJS entry point (which breaks the naming scheme but I don't think there's a way around making the Node.js loader accept a CJS in an ESM package).

PR #652, opened 2023-05-05
fix(package.json): add export map

christjt:
Running ESM in jest tests (node) isn't working properly with this lib yielding this error:
image

Running [@skypack/package-check](https://www.npmjs.com/package/@skypack/package-check) also validates this.
image

This PR adds export map and after these changes everything is healthy:
image

Package.json `"module"` is legacy and can (and probably should) be removed. You can also concider adding `"type": "module"` to properly indicate that this lib is ESM, but again, I'll leave this to the maintainer to decide.

This may also fix: https://github.com/tweenjs/tween.js/issues/649

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,11 +7,11 @@
 	"types": "dist/tween.d.ts",
 	"module": "dist/tween.esm.js",
 	"exports": {
-    		".": {
-      			"import": "./dist/tween.esm.js",
-      			"require": "./dist/tween.cjs.js"
-    		}
-  	},
+		".": {
+			"import": "./dist/tween.esm.js",
+			"require": "./dist/tween.cjs.js"
+		}
+	},
 	"files": [
 		"dist",
 		"README.md",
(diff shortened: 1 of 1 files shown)


trusktr:
This is great! I do agree we should change to type:module and remove the old field.

We would probably have to release this as a major bump, along with a release note too I think.

Would love to get rid of CJS, and just have ESM.

christjt:
> This is great! I do agree we should change to type:module and remove the old field.
> 
> We would probably have to release this as a major bump, along with a release note too I think.
> 
> Would love to get rid of CJS, and just have ESM.

That all sounds great, but we we can add the exports field without it being a breaking change, right? And then in the future considering to remove the commonjs output if you want.

christjt:
@trusktr

trusktr:
@christjt Hello! I'm in favor of this change for now. I think I would prefer to just ship vanilla ES Modules (no bundle) and let people bundle any way they want (or just import vanilla). Simplifies the setup.

For some reason the build is not running. Do the regular tests pass for you? Can you also add the skypack check to the tests?

trusktr:
Ok, got the darn build fixed. Looks like it was a bug inside GitHub.

trusktr:
Ok, tests passing, gonna just merge and release a major bump!

trusktr (approved):

PR #659, opened 2023-06-23
Fix CJS entry point

### LINKED ISSUE #649: CJS entry point can't be loaded in Node.js

inukshuk:
Closes #649


pailhead:
I am unable to import TWEEN with typescript, whatsoever.

SilkePilon:
this needs to be merged rn.....

trusktr:
Hello! Is this still needed after the recent changes in `main`?

The `exports` field now includes the link to the `types`, however the cjs file name has not changed.

Can you import in TypeScript now @pailhead?

Do we still need to change the name @inukshuk?

trusktr:
Yep, the naming needed to be changed. I added an example Node.js CommonJS project in `examples/example-projects/nodejs-commonjs/` to test it out.

Thanks! Merged!

trusktr:
Release in v22.0.0 (major versio bump because the file name changed, just in case anyone somehow relied on the `.cjs.js` file in any way).

Issue #655, opened 2023-06-23
Simplify Tween, remove some stuff, make it easier to maintain.


trusktr:
Proposed (welcome to debate):

- [ ] Remove the `TWEEN` singleton.
  - Teach people how to make a Tween and update it directly.
  - Teach them that if they want to update multiple Tweens at once, they can make a Group.
  - This eliminates possible errors people will have when they clash tweens from two or more separate pieces of code.
- [ ] Remove relative values. It involves detecting strings for little convenience. People can just use math on their side with not much more effort.
- [ ] Remove the dynamic-to feature. It isn't used often, and much internal complication.
- [ ] Remove array of to-values.
  - Also adds quite some complexity.
  - It seems that when we do this, we inadvertently destroy the meaning of an easing curve because the actual number output is no longer on that curve so the meaning is obscure, unless we do very special math to determine proper values to use such that a new curve makes sense.
  - The same math skill required for the previous point may instead be used to chain multiple tweens together, which gives more capabilities such as specific curves between each value. To emulate the array of values, chain tweens together with linear curves, then interpolate all of them with a single higher-level tween using a single curve.
  - The previous point allows much more options than the inflexible feature of to-values. To-values is a shortcut for the previous point with a linear chain and directing tween.
  - Downside: lose ability to choose interpolation method, f.e. bezier, which is useful for making a smooth curve out of the discrete to-values.
    - provide an alternative that doesn't complicate tween directly? F.e. Three.js has Path and Curve tools, instead animate position along such primitives specialized for that purpose. Perhaps we show a demo of that here, after removing to-values.

What else?

trusktr:
@MasatoMakino hello! :) If you have some time, we could jam on these ideas.

MasatoMakino:
@trusktr hello!

#### `TWEEN` singleton.

> docs: Teach them that if they want to update multiple Tweens at once, they must make a Group.

Does this mean the [mainGroup](https://github.com/tweenjs/tween.js/blob/e01bbdbebe2d05509b73bbbe7dc7438228124a9e/src/Tween.ts#L52C62-L52C62) is deprecated?

#### Remove the dynamic-to and array of to-values feature

I agree with the removal of dynamic-to and to-array. These are very large features. If necessary, I think they should be separated into independent npm packages or replaced with reliable packages.

I agree with the policy of duplicating tween in utility functions rather than adding functionality to tween instances. it makes `Tween.ts` easier to maintain.

#### Deprecated steps

[...]

PR #661, opened 2023-06-28
fix(package.json): fix export map types

tazyong:
Fix import error in typescript
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,14 +8,9 @@
 	"module": "dist/tween.esm.js",
 	"exports": {
 		".": {
-			"import": {
-				"types": "./dist/tween.d.ts",
-				"default": "./dist/tween.esm.js"
-			},
-			"require": {
-				"types": "./dist/tween.d.ts",
-				"default": "./dist/tween.cjs.js"
-			}
+			"import": "./dist/tween.esm.js",
+			"require": "./dist/tween.cjs.js",
+			"types": "./dist/tween.d.ts"
 		}
 	},
 	"files": [
--- a/package.json
+++ b/package.json
@@ -8,8 +8,14 @@
 	"module": "dist/tween.esm.js",
 	"exports": {
 		".": {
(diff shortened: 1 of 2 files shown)


bachirelkhoury (approved):
This resolved the typing error

trusktr:
This will be handled soon as part of

- https://github.com/tweenjs/tween.js/issues/655.

Besides it won't make sense to point both ESM and CommonJS files to the same types.

So in #655 we'll work on this next first thing, simplify the output, moving onto only ES Modules.

package.json
```diff
@@ -8,8 +8,14 @@
 	"module": "dist/tween.esm.js",
 	"exports": {
 		".": {
-			"import": "./dist/tween.esm.js",
```
july-12, line 11: could use more easy way by just one line inserted like this ?
image
it seems work for me!

DJ-caddev:
Any update on this issue?

MasatoMakino:
Here is a minimal tsconfig.json to reproduce this problem.

```json
{
  "compilerOptions": {
    "module": "ES2015",
    "noImplicitAny": true,
    "moduleResolution": "bundler" // or "node16"
  }
}
```
`noImplicitAny` is [part of `strict`](https://www.typescriptlang.org/tsconfig#strict).

I hope this record helps someone.

trusktr:
Thanks @MasatoMakino. Will circle to this soon, this week.

Methuselah96:
Hey @trusktr this is a pretty rough for TypeScript users, since TypeScript doesn't think there are any types for modern `moduleResolution` settings. I had to [copy-and-paste the types into `@types/three`](https://github.com/three-types/three-ts-types/pull/593) to work around this.

Looking forward to this getting resolved. :)

trusktr:
The types have been recently added from the other PR (the one-liner as @july-12 mentioned above)!

trusktr:
Thank you for the contribution!

PR #662, opened 2023-07-01
docs: update tutorial link

victorbnl:
The tutorial in the README now redirects to a page to buy Instagram likes. I tend to think that’s not the expected tutorial, so I replaced it with the archived version.


trusktr:
Ah, nice catch. Thanks!

PR #667, opened 2023-09-06
Add missing types file to package.json exports

jramstedt:
Fixes types loading when tsconfig moduleResolution is set to 'bundler'


monkpatch:
+1
I was about to make just the same pull request :D
Have this problem too :(

trusktr:
thank you!

Issue #668, opened 2023-10-07
when time is negative, the result is bad


jackletter:
see code(vue):
```js
let process2 = ref(1);
new Tween(process2)
.to({ value: 2 }, -2*1000)
.onUpdate((obj, el) => {
    console.log(`obj.value=${obj.value}`);
})
.onComplete(() => {
    console.log(`process2.value=${process2.value}`)
})
.start();
```
we can see console:
```html
......
Three6.vue:19 obj.value=0.01034999999962749
Three6.vue:19 obj.value=0.0022499999999999742
Three6.vue:19 obj.value=-0.006200000001117534
Three6.vue:19 obj.value=-0.014450000001117624
Three6.vue:19 obj.value=-0.022700000001117493
Three6.vue:19 obj.value=-0.031200000001117667
Three6.vue:19 obj.value=-0.039550000000745156
Three6.vue:19 obj.value=-0.04774999999999996
......
```

expect:
when time is negative, has no animate,just like time is 0

trusktr:
what's your use case? why animate to a negative time? can you avoid doing that?

jackletter:
收到了 哈哈

trusktr:
I'm not sure this is something we should worry about because there are lots of things you could pass into Tween API and it would break. Just pass a positive number as documented. If your system has negative numbers, just map them to positive.

Closing,  but if you feel this really needs to be fixed, feel free to reply, and/or make a PR.

trusktr:
Merged and released in 23.1.1

PR #669, opened 2023-10-08
fixbug:when duration is negative,there should has no animate,not a bad result

jackletter:
#668


trusktr:
Thanks! This is helpful!

PR #673, opened 2023-12-17
fix: Use type definition in ESM

keiya01:
I noticed type definition isn't used when I specify `"type": "module"` in package.json. So I fixed this issue.


kebkaldanil:
yep, works for me. Without it TS shows error

KaSSaaaa (approved):

trusktr (approved):

Issue #600, opened 2021-03-01
it would be nice to be able to get the duration we assign to a tween


cesarl94:

trusktr:
It's stored on this property:

https://github.com/tweenjs/tween.js/blob/ae24c58c8b57b2079151ae2c3c63f2bb501c8b2b/src/Tween.ts#L26

But it is marked as private if you're using TypeScript. No problem in plain JS though. It likely won't change.

I think we can do something special like `duration()` without an argument returning the current value, or something.

trusktr:
available in v23.1.0

PR #674, opened 2023-12-22
Feature: Add method getDuration to tween

kdb9520:
Implementing suggestion of issue #600.  Added getDuration method to return the private _duration attribute of a Tween.


trusktr:
Thanks!

Issue #677, opened 2024-01-25
[bug] yoyo has wrong value


alphabetabc:
# demo
```ts
const rootEl = document.getElementById("root");
Object.assign(rootEl.style, {
    width: "800px",
    height: "600px",
});

rootEl.innerHTML = "loading";

const dataX = [];
const dataY = [];

let i = 0;
createAnimationQueue({
    from: { value: 1 },
    to: { value: 1.5 },
    yoyo: true,
    repeat: 3,
    onUpdate: (o) => {
        dataX.push(i++);
        dataY.push(o.value);
    },
    onComplete: () => {
        const chart = ec.init(rootEl);
        chart.setOption({
            xAxis: {
                type: "category",
                data: dataX,
            },
            yAxis: {
                type: "value",
            },
            series: [
                {
                    data: dataY,
                    type: "line",
                },
            ],
        });
    },
});

```

# result
![image](https://github.com/tweenjs/tween.js/assets/29876424/2c4990e7-15bd-40db-9c45-8dcfa22a4c3f)

mk965:
Hello, can you provide a link to the [codepen](//codepen.io) containing the complete demo?

humodz:
https://codepen.io/humodz/pen/YzgEpXG

When the red ball reaches the end of the animation, on both directions, it blinks to the opposite side for a split second

humodz:
As a workaround until it's fixed, we can call TWEEN.update twice, like so:

```ts
requestAnimationFrame(function animate(time) {
  requestAnimationFrame(animate);
  TWEEN.update(time);
  TWEEN.update(time);
});
```

This twill trigger callbacks twice which is a bummer, but at least the weird jump goes away

trusktr:
Fixed in 

- https://github.com/tweenjs/tween.js/pull/683 thanks to @humodz !

Released in v23.1.2

trusktr:
Re-opening, as the merge for this broke the yoyo demo, which I've reverted. See

- https://github.com/tweenjs/tween.js/issues/687

We assumed passing unit tests were enough, but we forgot to test the actual yoyo demo! 🥴

bendytree:
Fwiw - I'm having the same issue on both `23.1.3` and `25.0.0`. @humodz's easing workaround worked for me.

ManHunter:
Is anyone working on a fix? this is a pretty big bug that makes it impossible to use yoyo.

FrostKiwi:
[...]

PR #678, opened 2024-02-12
fix: when yoyo repeats, first frame behaves as if yoyo=false

humodz:
fix for https://github.com/tweenjs/tween.js/issues/677

summary:

The current logic first updates the properties, then run the logic that reverses the start and end values for yoyos. The problem is that if the tween has just started repeating, the update will happen before the reversing, so for a single frame the value will be wrong, as if yoyo=false

The proposed fix is to check if the tween has just started to repeat and, if so, apply the yoyo logic before updating the properties

Also, something that I noticed, but I'm not sure if it's intended or not: when the tween repeats, onEveryStart seem to trigger one update later than I would expect. For example:

```ts
new TWEEN.Tween(obj)
    .to({ x: 100 }, 100)
    .repeat(1)
    .yoyo(true)
    .start(0);

TWEEN.update(99); // start everystart update
TWEEN.update(120); // update repeat
TWEEN.update(150); // everystart update
TWEEN.update(199); // update
TWEEN.update(201); // update complete
```

shouldn't everystart happen at 120ms, since the tween has already repeated at that instant?

### DIFF
--- a/dist/tween.amd.js
+++ b/dist/tween.amd.js
@@ -678,13 +678,11 @@ define(['exports'], (function (exports) { 'use strict';
          * it is still playing, just paused).
          */
         Tween.prototype.update = function (time, autoStart) {
-            var _this = this;
             var _a;
             if (time === void 0) { time = now(); }
             if (autoStart === void 0) { autoStart = true; }
             if (this._isPaused)
                 return true;
-            var property;
             var endTime = this._startTime + this._duration;
             if (!this._goToEnd && !this._isPlaying) {
                 if (time > endTime)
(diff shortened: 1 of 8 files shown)


alphabetabc:
I found that this code can be fixed by restoring it to this state
image

humodz:
> I found that this code can be fixed by restoring it to this state image

This fails if `.update` is called with a number larger than the repeat interval, so that multiple repetitions would elapse on a single update

trusktr:
@humodz thank you for this excellent and well-tested work! For some reason the tests are not running, so I move your code into a local branch and make a new PR with the same commits:

https://github.com/tweenjs/tween.js/pull/683

Will close this one in faavor of that one.

trusktr:
@humodz for some reason this broke the yoyo example:

- https://github.com/tweenjs/tween.js/issues/687

so I had to revert the change (which I had originally merged in https://github.com/tweenjs/tween.js/pull/683).

Although all tests were passing, something was different with the yoyo example that we will need to account for in the unit tests.

PR #682, opened 2024-05-02
Change the Easing.Linear methods to pure function

soumen-pradhan:
The methods `Linear.In`, `Linear.Out`, `Linear.InOut` reference the method `Linear.None`. Saving a reference to any variable and calling them will throw an error, since `this` context is no longer the same. This is not an issue with any of the other `Easing` methods.

```js
Easing.Linear.In(5) // 5

let f = Easing.Linear.In
f(5) // Uncaught TypeError: Cannot read properties of undefined (reading 'None')
```

None of the other methods refer to `this` in their body, and can be called directly or through a variable reference.

So, I have replaced `this.None` to simply the body of `Linear.None`. A trivial change, should not break anything.


trusktr (approved):

trusktr:
Nice, thanks!

PR #685, opened 2024-05-12
Tiny readme fix for import-map example code.

Bug-Reaper:
Corrects example so the import-map path is relative.

> Copy and paste current import-map example code it's gonna trigger an error:
> Screenshot 2024-05-12 at 7 05 09 PM

Also changes `tween.es.js` => `tween.esm.js`. There does not seem to be any `tween.es.js` available in the default npm install so assume this is a typo.

---

The importmap example code works for me after these two small fixes ✌️

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ You can import from `node_modules` if you serve node_modules as part of your web
 <script type="importmap">
 	{
 		"imports": {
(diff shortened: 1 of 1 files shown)


trusktr:
Thank you!

Issue #697, opened 2024-12-18
Make it possible to change "now"


J00nz:
As a developer it would be very nice to have a way to override the internal "now" functionality.

I realize that it is possible to send it, or rather "time2, as a argument when invoking methods, but that is a very cumbersome way of doing it.

I propose a way to simply "setNow" that takes a function as a argument and overrides "now" to this.

It would make implementations that modifies the "rate" of time much more convenient to do.

linorabolini:
This is golden. Any clues on when it would be published ?

trusktr:
Sorry for the late reply. 

The .start() and .update() methods are how to set the time value with custom values to do any animation of time as needed. The tests show examples.

If a use case is not covered, open a new issue to describe with code what is needed.

PR #698, opened 2024-12-18
#697: Added "setNow"

J00nz:
Added the ability to override the internal "now" function using the "setNow" function.

It simply takes the new function you want to override the default "performance.now" implementation.

Useful when you want to control the flow of time on a global scale in a dynamic manner.

### DIFF
--- a/dist/tween.amd.js
+++ b/dist/tween.amd.js
@@ -215,7 +215,13 @@ define(['exports'], (function (exports) { 'use strict';
         },
     });
 
-    var now = function () { return performance.now(); };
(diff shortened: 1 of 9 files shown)


trusktr:
Hello @J00nz, looks good! Can you mention this in the guide? Then it will be good to go.

trusktr:
I'll try to get to it tomorrow if you haven't already and then we'll get it in. :)

J00nz:
Great!
I added a section in the user-guide, I do however not write chinese, so perhaps it would be good to get help with that 🙏

trusktr (approved):

PR #709, opened 2025-09-11
fix(yoyo): fix yoyo wrong -- 20250911

alphabetabc:


trusktr:
Hello! thanks for attempting a change here. Tween.yoyo (and .delay/.repeat) will be deprecated, and eventually removed, replaced with a new `Timeline` feature for such orchestrations.

- #713

PR #712, opened 2026-08-31
Refactor: Optimizations, cleanup and chore improvements

dalisoft:
## Summary

A toolchain and housekeeping pass over the repository: the project moves from
Node.js/npm + rollup + prettier + nodeunit to Bun + Biome, the committed build
artifacts are dropped, and the sources, examples and docs are cleaned up and
modernised accordingly.

73 files changed, +3645 / −17447. The unit suite (109 tests, 871 assertions)
passes.

## Toolchain

- **Runtime / package manager → Bun.** `packageManager` is pinned to
  `bun@1.4.0`, `package-lock.json` is replaced by `bun.lock`, and Bun is now
  the only prerequisite for development (no separate Node.js/npm install).
- **Bundler: rollup → `Bun.build`.** `rollup.config.js` and its plugins are
  gone, replaced by `scripts/build.ts`, which emits:

  | Output | Format | Target |
  | --- | --- | --- |
  | `dist/tween.esm.js` | esm | node |
  | `dist/tween.cjs` | cjs | node |
  | `dist/tween.umd.js` | iife | browser |
  | `dist/tween.umd.min.js` | iife, minified | browser |

  The AMD bundle is dropped — it was never referenced from `main`/`exports`.
  Filenames are unchanged, so existing `<script src>` and CDN URLs keep
  working; `dist/tween.umd.min.js` is new (11.8 kB raw / 3.8 kB gzip).
- **One build command.** `scripts/build.ts` bundles, then invokes
  `tsc --project tsconfig.build.json` to emit declarations into `dist/types`,
  then writes `dist/tween.d.ts` re-exporting the entry point so the published
  `types` path is unchanged. `bun run build` is the whole pipeline.
- **Browser entry point.** Bun's bundler has no UMD format, so `src/global.ts`
  stands in for what rollup's `format: 'umd'` produced — an iife bundle that
  assigns `TWEEN` onto `globalThis`.
- **Linter/formatter: prettier (+ eslint comments) → Biome.**
  `.prettierrc.cjs` / `.prettierignore` are replaced by `biome.json`, which
  keeps the existing style (tabs, single quotes, no semicolons, no bracket
  spacing, 120 columns) so diffs stay readable. The HTML linter is enabled,
  `a11y/useMediaCaption` is lowered to `info` and
  `suspicious/noAssignInExpressions` to `warn`, and the vendored
  `examples/js/stats.min.js` plus the generated JS under
  `examples/example-projects/plain-typescript-modules` are excluded globally
  via `files.includes` rather than per-tool.
- **Markdown lint: `rumdl`,** configured in `.rumdl.toml` with
  prettier-compatible rules.
- **Tests: nodeunit → `bun:test`.** `src/tests.ts`,
  `src/test-performance-now-fake.ts`, the vendored `test/unit/nodeunit.js`
  (~2.2k lines), `nodeunit.html` and `nodeunitheadless.cjs` are replaced by
  `test/unit/tween.test.ts` and `test/unit/performance-now-fake.ts`. Tests now
  live under `test/` rather than in `src/`, so they no longer ship in the
  compile graph of the library.
- **Git hooks: `lefthook`.** `pre-commit` runs Biome and rumdl on staged
  files; `pre-push` additionally runs `tsc --noEmit` and `bun test`.
- **TypeScript 5.0.4 → 7.x.** `tsconfig.json` drops the deprecated
  `importsNotUsedAsValues` / `ignoreDeprecations` pair, switches to
  `module: preserve` + `moduleResolution: bundler`, raises `target`/`lib` from
  `es5` to `es2022`, and is `noEmit` by default (emitting is
  `tsconfig.build.json`'s job).
- **CI** now uses `actions/checkout@v7` + `oven-sh/setup-bun@v2` with the Bun
  version read from `package.json`, and runs
  `bun install --frozen-lockfile` → `bun run lint` → `bun test`. The Node
  version matrix is removed.

## Package metadata

- **`engines.node` is now declared: `>=22.12`.** The package had no `engines`
  field before. 22.12 is the oldest Node.js release that is both still
  supported (Node 20 reached EOL on 2026-04-30) and able to `require()` an ES
  module natively.
- **Entry points:** `main` and `module` point at the ES module build, `types`
  at `dist/tween.d.ts`, and `exports` lists `types` / `import` / `require` in
  that order. The iife bundle is deliberately *not* exposed as an `exports`
  condition: it has no module exports, so any resolver that picked it would
  hand the consumer an empty namespace.
- **CDN default files.** `jsdelivr` and `unpkg` point at
  `dist/tween.umd.min.js`, and `browser` at `dist/tween.umd.js`, so a
  bare `https://cdn.jsdelivr.net/npm/@tweenjs/tween.js` or
  `https://unpkg.com/@tweenjs/tween.js` serves the global build ready for a
  `<script>` tag. Bundlers are unaffected: `exports` takes precedence over all
  three fields.
- **`files`** is narrowed to `dist`, `README.md` and `LICENSE`.

## Repository hygiene

- `dist/` is no longer committed (~6.4k lines of generated code removed) and
  is now git-ignored; it is built on `prepublishOnly`, so published packages
  are unaffected.
- npm scripts are trimmed to what Bun needs: `build`, `examples`, `lint`,
  `prepublishOnly`, `version` and the three `release:*` scripts. `dev`,
  `tsc`, `tsc-watch`, `rollup-build`, `test`, `test-unit`, `test-lint` and
  `prepare` are dropped — `bun test` and `bun run lint` cover them directly.
- Version bumped to `25.0.1` (`package.json` and `src/Version.ts`, the latter
  generated by `scripts/write-version.ts`).

## Source cleanup

- All `eslint-disable` and `@ts-ignore`/`FIXME` suppressions in `src/Tween.ts`
  are removed; the code now type-checks without them.
- `isNaN` / `isFinite` → `Number.isNaN` / `Number.isFinite` (no coercion of
  non-numeric input).
- `Math.pow(x, y)` → `x ** y` throughout `src/Easing.ts`.
- Function expressions → arrow functions in `src/Interpolation.ts`, and the
  loop variable in `Tween.update` is scoped to the loop (`const property`)
  instead of being hoisted to the top of the method.
- `src/Now.ts` drops the unsafe `Function` type: the internal clock is typed
  `() => DOMHighResTimeStamp`, `now()` and `setNow()` are arrow/typed
  functions returning `number`.
- `src/Sequence.ts` becomes a plain object literal instead of a class with a
  single static method and static field.
- Relative imports inside `src/` carry explicit `.js` extensions. TypeScript
  copies import specifiers verbatim into the emitted `.d.ts`, so without them
  the published declarations fail to resolve for consumers using
  `moduleResolution: node16` / `nodenext` (TS2834).

## Docs and examples

- `README.md`, `README_zh-CN.md`, `docs/user_guide*.md` and
  `docs/contributor_guide*.md` are reformatted with rumdl and updated for the
  new workflow: `npm install`/`npm test` → `bun install`/`bun test`, prettier
  + eslint → Biome, nodeunit → Bun's test runner, and the test file path is
  corrected to `test/unit/tween.test.ts`. The build output list also gains the
  new minified UMD bundle. The stale note about the broken
  `test/unit/nodeunit.html` browser runner is dropped along with the file.
- Examples are reformatted by Biome and modernised: arrow-function callbacks,
  and string concatenation replaced by template literals throughout the
  example scripts and `examples/js/{createGraph,createPath}.js`.
- Example markup is corrected: `19_end_tween.html` gains the missing
  `<!doctype>`, `<html lang>`, `<head>` and `<title>`, and every `<button>`
  gets an explicit `type="button"` so it no longer defaults to submit
  behaviour.
- `examples/10_yoyo.js` no longer assigns to `window.*` inside declarations;
  the control functions are declared normally and exposed in one
  `Object.assign(window, Object.freeze({...}))` at the end.
- The stylesheets (`assets/style.css`, `examples/css/style.css` and the two
  example-project stylesheets) use the valid `sans-serif` generic family
  instead of the non-existent `sans`.

## Notes for reviewers

Consumer-visible changes, all of which argue for releasing this as a major:

- `target` moves from `es5` to `es2022`, so the shipped bundles are no longer
  downlevelled.
- `engines.node` is `>=22.12` where previously nothing was declared.
- `dist/tween.amd.js` is removed. It was never referenced from
  `main`/`exports`, and the AMD loaders it targeted are long out of use.
  Every other bundle keeps its filename, so `<script src>` tags, pinned CDN
  URLs and the cdnjs entry (which hardcodes `tween.umd.js`) all keep working.

The library's runtime API itself is untouched, and behaviour is covered by the
same 871 assertions as before.

`unsignZero` in the test file exists because nodeunit's `equal` used `==`
(which treats `-0` and `0` as equal) while `expect().toBe` uses `Object.is`;
`Circular.InOut(0)`, `Back.In(0)` and `Back.InOut(0)` return `-0`, so the sign
is normalised to preserve the original assertions.


dalisoft:
## Measured impact

Numbers below were produced by checking out `main` (20079e6) and this branch
(2cb2639) into two clean worktrees, installing from scratch in each, and running
the equivalent commands. Median of repeated runs; wall-clock via `/usr/bin/time -p`.

Environment: Apple M1 Max, macOS 15.7.9 (arm64), Bun 1.4.0, Node v26.8.1, npm 12.0.2.

### Install

| | `main` (npm) | this PR (Bun) | Δ |
| --- | ---: | ---: | ---: |
| Cold cache (fresh CI runner) | 20.35 s | 2.03 s | **10.0× faster** |
| Warm cache (local re-install) | 3.94 s | 0.40 s | **9.9× faster** |
| Direct devDependencies | 7 | 5 | −29% |
| Installed packages | 227 | 12 | **−95%** |
| Lockfile | `package-lock.json`, 4990 lines | `bun.lock`, 126 lines | −97% |

### Build

| | `main` (tsc + rollup) | this PR (`Bun.build` + tsc) | Δ |
| --- | ---: | ---: | ---: |
| Full build from clean | 1.47 s | 0.34 s | **4.3× faster** |
| Typecheck only (`tsc --noEmit`) | 0.82 s (TS 5.0.4) | 0.32 s (TS 7.0.2) | **2.6× faster** |

The 0.34 s covers everything: four bundles (esm, cjs, umd, minified umd),
declaration emit into `dist/types`, and the `dist/tween.d.ts` re-export.

### Lint / format

| | `main` (Prettier) | this PR (Biome) | Δ |
| --- | ---: | ---: | ---: |
| Check whole repo | 1.24 s | 0.12 s | **10.3× faster** |
| Tool-reported time | — | 43 ms / 63 files | — |

### Tests

| | `main` (nodeunit) | this PR (`bun test`) | Δ |
| --- | ---: | ---: | ---: |
| Run suite (already built) | 0.47 s | 0.03 s | **15× faster** |
| Run suite from a clean tree | 1.94 s (build required first) | 0.03 s (runs TS directly) | **65× faster** |
| Tests / assertions | 109 / 871 | 109 / 871 | parity |
| Test infrastructure LOC | 2838 (`src/tests.ts`) + 2174 (vendored `nodeunit.js`) + 18 | 2697 + 40 | −45% |

Test coverage is carried over 1:1 — same 871 assertions, no test dropped or weakened.

### Full CI pipeline

| | `main` | this PR | Δ |
| --- | ---: | ---: | ---: |
| Install + verify (cold cache) | 20.35 s + 2.94 s = **23.3 s** | 2.03 s + 0.15 s = **2.2 s** | **10.6× faster** |

`main` ran `npm clean-install` then `npm test` (build → lint → unit); this PR runs
`bun install --frozen-lockfile` then `bun run lint` → `bun test`. Removing the Node
version matrix also cuts the job count.

### Local developer loop

The new `lefthook` gates cost effectively nothing, so they can run on every commit/push:

| Hook | Contents | Time |
| --- | --- | ---: |
| `pre-commit` | Biome + rumdl on staged files | ~0.2 s |
| `pre-push` | typecheck + Biome + rumdl + full test suite, in parallel, over all files | **0.58 s** |

### Published artifacts

Bundle sizes, raw and gzip (`gzip -9`):

| File | `main` raw | PR raw | Δ raw | `main` gzip | PR gzip | Δ gzip |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `tween.esm.js` | 44,403 B | 22,807 B | **−48.6%** | 7,225 B | 4,775 B | **−33.9%** |
| `tween.cjs` | 44,725 B | 24,226 B | **−45.8%** | 7,335 B | 5,275 B | **−28.1%** |
| `tween.umd.js` | 50,726 B | 25,141 B | **−50.4%** | 7,566 B | 5,098 B | **−32.6%** |
| `tween.umd.min.js` | — | 11,816 B | new | — | 3,828 B | **−49.4%** vs `main`'s `tween.umd.js` |
| `tween.amd.js` | 50,433 B | — (dropped) | −100% | 7,459 B | — | −100% |
| `tween.d.ts` | 18,419 B | 68 B + `dist/types/*` (19,623 B) | ≈ parity | 2,811 B | 3,097 B | ≈ parity |

Package as published (`npm pack --dry-run`):

| | `main` | this PR | Δ |
| --- | ---: | ---: | ---: |
| Tarball | 36,347 B | 20,337 B | **−44.0%** |
| Unpacked | 228,051 B | 121,807 B | **−46.6%** |
| Files | 8 | 17 | +9 (per-module `.d.ts` + minified bundle) |

Most of the size reduction comes from the `es5` → `es2022` target: no more
downlevelled classes, spread helpers or `tslib`-style shims. A script-tag
consumer that switches to `tween.umd.min.js` downloads 3.8 kB gzip instead of
7.6 kB — half the bytes of the old UMD bundle.

### Repository

| | `main` | this PR | Δ |
| --- | ---: | ---: | ---: |
| Tracked files | 128 | 123 | −5 |
| Tracked bytes | 1,571,777 B | 1,151,633 B | **−26.7%** |
| Diff | | | +3,645 / −17,447 lines |

`dist/` is no longer committed, so builds stop producing review noise and merge
conflicts in generated files.

### Runtime performance

The library's own hot path was benchmarked directly: 2000 tweens × 600
`group.update()` frames, easing `Quadratic.InOut`, 7 iterations per process
after warmup, repeated across 4 fresh processes per configuration.

| `dist/tween.esm.js` | Node 26 (V8) | Bun 1.4 (JSC) |
| --- | ---: | ---: |
| `main` | 9.91 M updates/s (9.74–10.14) | 7.37 M updates/s (6.77–7.61) |
| This PR | 10.07 M updates/s (9.92–10.11) | 7.22 M updates/s (7.04–7.26) |
| Δ (median) | +1.6% | −2.0% |

Both deltas sit inside the ±4% process-to-process spread of the benchmark
itself, so throughput is unchanged: the `es5` → `es2022` bundle neither costs
nor gains measurable performance on this workload. Cold module import is
marginally cheaper (1.08 ms → 0.99 ms under Node), consistent with the smaller
bundle.

Issue #647, opened 2023-04-23
`Timeline` feature


trusktr:
The `Tween` class is doing a lot. Maybe it is starting to grow out of scope for what a Tween is. For example, we can `chain`, `yoyo`, `repeat` tweens, etc. Maybe it would be better to have a cleaner separation of concerns and introduce a new `Timeline` class (maybe convert `Group` with `Timeline`) that would be generally more robust for this.

What could this `Timeline` API looks like? How would `Tween` play with `Timeline`? Perhaps `Tweens` are added to `Timelines`.

This comment has some ideas: https://github.com/tweenjs/tween.js/issues/560#issuecomment-752838191

Food for thought:

```js
const track1 = new Track([
  new Empty(300),
  new Tween(mesh.rotation).to(..., 1000),
  new Tween(mesh.position).delay(500).to(..., 1000),
])

const track2 = new Tween(otherThing).to(2000) // considered as a track by itself

const animation = new Timeline([
  // tracks for the timeline
  track1,
  track2
]).start() // some Timeline API is shared with Tween

requestAnimationFrame(function loop() {
  animation.update()
  requestAnimationFrame(loop)
})
```

cuixiping:
`Group` and `.chain()`  can be enhanced to cover the cases.
`Group` should have all methods of tween instance.
`.chain()` should can be chained more times.

linorabolini:
I'd say that the most beautiful thing about tween.js is that one needs to learn just 2 things. Group and Tween. 

It's quite simple to just create a sequence with a simple `for` loop and just subsequent chains.
I made a fork to play on my own and I tried adding a static Sequence function to Tween.

```
const firstTween = Tween.Sequence(t1, t2, t3, t4);
```

I didn't like it much. It feels like a nice-to-have in the lib but also an extra. Moving it to Group would be too complex because groups don't have a notion of the added order of the tweens.

I have to admit that I added a private _lastAddedTween reference to Group so as to be able to have a queue of tweens, using Group as a "timeline"

```
group
.addQueued(tween1) // queue empty, auto-start tween1
.addQueued(tween2) // chain tween1 to tween2 
.addQueued(tween3) // chain tween2 to tween3
.addQueued(tween4) // chain tween3 to tween4
```

It works quite quite well, but I can't stop thinking that this could be done as a separated thing and not as part of the based group/tween lib.

Issue #560, opened 2020-07-29
Clock synchronization issue


christopherjbaker:
When controlling the clock by passing a value to `TWEEN.update(...)`, the default `_startTime` for a tween started with `.start()` still uses the browser clock, which means tweens do not behave as expected. I've fixed this in my own code by redefining `TWEEN.now` to return the last value I sent to `.update(...)`. It seems to me that if its in a "controlled mode", everything should use those controlled values.

trusktr:
That change may work for your case, but I think it may break individually-controlled Tweens. `TWEEN.now()` is merely supposed to be an API in place of `performance.now()` in case some browser doesn't have that.

If interested to make a PR but adding this feature behind an API more like `TWEEN.lastTime()` or similar, and update the tests, that would be great!

trusktr:
Do you have a code snippet that demonstrates what issue you had?

christopherjbaker:
I don't have a code snippet, but here is how you can recreate it:
1. Set up a RAF that calls `TWEEN.update(Date.now() + 2500)`. This sets up the internal clock so that it will _not_ line up with the browser's clock. In my case, I'm using threejs's clock, which tracks in seconds instead of ms.
2. Set up an animation that will take 5 seconds to complete. Do not pass any arguments to `start()`.
3. Observe that it does not behave as intended. I believe in this case, it will start halfway done, because it "started" at the current time, but the update was called with the current time + 2.5 seconds.

You've provided control of the internal clock, but you don't use the custom clock values when you start a tween, so they start in a different timeframe than the updates.

trusktr:
Interesting, haven't used it like that before. Normally what I do is control every tween individually, which may be a workaround for now:

```js
const tween = new Tween(...)....start(startTime)
// ...
tween.update(newTime)
```

That way each tween can operate with its own custom time being passed in (plus easier to encapsulate the tween to a particular component of your UI, pause it individually, update it individually, etc, while other tweens in other components can also do their own thing).

audiodude:
I'm going to try and write a PR that addresses this, because I've been bitten by it too when trying to programmatically step through time while using tween.js.

audiodude:
Here's the basic idea of what I'm trying to do: https://jsfiddle.net/audiodude/anL0tzux/32/

audiodude:
[...]

PR #713, opened 2026-09-11
feat!: Timeline orchestration (sequence/parallel, nesting, labels)

dalisoft:
Implements the Timeline feature from #647 (per @trusktr's direction in #560/#710).

## Design: Tween tweens, Timeline orchestrates

Legacy timing methods (`delay`/`repeat`/`repeatDelay`) are deprecated on `Tween` (still working, removal in a future major). `Tween.yoyo()` is deprecated (use Timeline composition). `Timeline` replaces them; it has no timing methods of its own.

What changed (latest commits on top of placement-API rework):
- feat(tween): `clone()`, disable `yoyo()` (throws), defer `reverse()` (throws stub with typings)
- feat(timeline): `{repeat}` expansion, clone-on-readd, `Timeline.clone()` (`{yoyo}` option present but throws until `reverse()` lands)
- test: clone/throws/expansion tests (1050 assertions OK)
- docs + examples (20 explicit forward/back clips, 21 slider), dist rebuilds

Features:
- Placement (typed, no string DSL): `add(x)` appends, `add(x, 500)`, `add(x, 'label')`, `add(x, otherTween)`, `add(x, {at, atIndex, offset, shift})`. Labels `start`/`end` builtin and protected.
- Repeats: `timeline.add(tween)` x3, or shortcut `add(tween, {repeat: 3})` (total plays; extra plays clone internally, each clip independent; clones snapshot start values immediately so repeats replay identically).
- Yoyo: deferred. `Tween.reverse()` throws until implemented; then `add(tween)` + `add(tween.reverse())` and `add(tween, {yoyo: true})` will work (expansion code already in place).
- Nesting, pause/resume/stop, onStart/onEveryStart/onUpdate/onComplete/onStop, easing/interpolation fan-out. Timeline can sit in a Group.

How it works: children run on a local clock (0 = timeline start) and start lazily when the playhead reaches them (one ever-set flag per entry, no extra collections), so sequential same-property tweens chain end-to-start. Scrubbing re-enters dead children within range; lower-clamped drive snaps to start values.

Usage:

```js
import {Timeline, Tween} from '@tweenjs/tween.js'

const seq = new Timeline()
  .add(new Tween(a).to({x: 100}, 500))
  .add(new Tween(b).to({y: 100}, 500)) // replaces chain()

const par = new Timeline()
  .add(new Tween(a).to({x: 100}, 1000), 0)
  .add(new Tween(b).to({y: 100}, 1000), 0)

const tl = new Timeline()
  .add(tweenA, 0)
  .addLabel('mid', 300)
  .add(tweenB, {at: 'mid', offset: 150})
  .add(tweenC, tweenB) // align to tweenB's start

const rep = new Timeline().add(tween, {repeat: 3})

// manual scrub (slider): restart at 0 when finished, then drive
if (!tl.isPlaying()) tl.start(0)
tl.update(value)

function animate(time) {
  requestAnimationFrame(animate)
  seq.update(time)
}
```

Demos: examples/20_timeline.html, examples/21_timeline_slider.html (ESM importmap; serve via `npm run examples`, see examples/README.md). Docs: Timeline section in docs/user_guide.md.

Conflicts: none with existing Tween/Group flows (Tween timing internals untouched). Don't put Timeline children in a Group too (double update). Sequence id-generator left untouched.

Group types widen Tween -> Tween|Timeline. A group can now update tweens and timelines.

Future (separate PRs): `Tween.reverse()` implementation (+ `{yoyo}` lights up), `timeline.undo()` state tracking for UI.


examples/20_timeline.html
trusktr, line 1: Let's use only ESM in new examples. UMD is legacy, and will be removed in a future version. Use an importmap script to map the package name to the file path of the module entry point in dist/.

docs/user_guide.md
```diff
@@ -320,6 +320,47 @@ Remove all tween from a group.
...
+```
+
+Positions also accept labels: `'myLabel'`, `'myLabel+=100'`, `'<'` (start of
+the last child), `'>'` (end), `'+=100'` (relative to the end). Custom labels
+are added with `timeline.addLabel('intro', 300)`. Timelines can be nested
+inside other timelines, and have their own `delay`, `repeat`, `repeatDelay`,
+`yoyo`, `pause`/`resume`, `stop`, and `onStart`/`onEveryStart`/`onUpdate`/
+`onRepeat`/`onComplete`/`onStop` callbacks, mirroring `Tween`. A `Timeline`
+can also be added to a `Group` (`group.add(timeline)`) — just don't add its
+children to a `Group` as well, or they would be updated twice.
```
trusktr: On line 353 to 360 the  paragraph is too dense. Improve the format using list items to better show the examples. Make sure all APIs are listed.

examples/21_timeline_slider.html
```diff
@@ -0,0 +1,169 @@
+<!doctype html>
+<html lang="en">
+	<head>
+		<title>Tween.js / timeline slider</title>
+		<meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+		<link href="css/style.css" media="screen" rel="stylesheet" type="text/css" />
+		<!-- UMD build (not a module) so this page also works via file:// -->
+		<script src="../dist/tween.umd.js"></script>
```
trusktr: As with the other comment, let's use JS modules with these examples.

src/Version.ts
```diff
@@ -1,2 +1,2 @@
-const VERSION = '25.0.0'
+const VERSION = '26.0.0'
```
trusktr: Let's undo the version change. We will do that during a release. Also this will be a non-breaking change so we'll update the minor once it is ready.

trusktr (changes requested):
@copilot see comments and update.

trusktr:
> * add() appends by default (replaces chain); add(x, 0) for parallel; explicit offsets, labels (afterInit/afterLast, custom, <, >, +=, label+=)

@copilot explain this

examples/20_timeline.html
```diff
@@ -17,6 +17,10 @@
 			<h1><a href="http://github.com/tweenjs/tween.js">tween.js</a></h1>
 			<h2>20 _ timeline</h2>
 			<p>Sequencing with Timeline: sequential by default, parallel via offsets and labels.</p>
+			<p>
+				<strong>Note:</strong> This ESM example should be served over HTTP, for example with
+				<code>npm run examples</code>.
+			</p>
```
trusktr: Put this in a README.md or similar file, not in the example. Update both examples.

trusktr:
> * `add(x, '<')` means align to the previous child’s start.

@copilot What is "previous child" here?

docs/user_guide.md
```diff
@@ -320,6 +320,85 @@ Remove all tween from a group.
...
+
+Timeline positions and APIs:
+
+- Position strings accepted by `timeline.add(...)` and `timeline.addLabel(...)`:
+  - `'myLabel'`
+  - `'myLabel+=100'`
+  - `'myLabel-=100'`
+  - `'<'` (start of the last child)
+  - `'>'` (end of the timeline)
+  - `'+=100'` / `'-=100'` (relative to the end)
```
trusktr: Lines 357-362: we don't want string parsing for a custom language.

Replace this with a well-typed API.

For example, 

```ts
timeline.add(tween, {at: 'myLabel', offset: 100})

timeline.add(tween, {at: 'myLabel', offset: -100})

timeline.add(tween, {at: myOtherTween, offset: 100})

const existingTween = timeline.getAll()[5]
timeline.add(tween, {at: existingTween, offset: -100})

// shortcut for the previous example. `atIndex` takes precedence over `at`.
timeline.add(tween, {atIndex: 5, offset: -100})

// place the tween starting at time 500-100
timeline.add(tween, {at: 500, offset: -100})

// when shift is true, all items in the timeline are shifted in time so that nothing  overlaps with the inserted tween.
timeline.add(tween, {atIndex: 5, shift: true})

// this shifts everything by duration of the inserted tween, and the timeline now starts with the inserted tween:
timeline.add(tween, {at: 0, shift: true})
```

Rename the default "afterInit" and "afterLast" labels to "start" and "end". Ensure that the user cannot modify start or end labels with `addLabel`. They can use their own custom labels instead of modifying "start" for example.

No more "<" or ">" special labels. Delete.

The two-arg form of timeline.add() will still exist. The second arg can be a tween, a time value, or a label string, and if they need more control such as specifying both a label and an offset, or options like shift, they can pass in an options object.

docs/user_guide.md
```diff
@@ -320,6 +320,85 @@ Remove all tween from a group.
...
+  - `timeline.start(time?)`
+  - `timeline.update(time?)`
+  - `timeline.isPlaying()`
+  - `timeline.isPaused()`
+  - `timeline.getDuration()`
+  - `timeline.getTotalDuration()`
+  - `timeline.delay(ms)`
+  - `timeline.repeat(times)`
+  - `timeline.repeatDelay(ms?)`
+  - `timeline.yoyo(true)`
```
trusktr: Lines 379-382:

The delay(), repeat(), repeatDelay(), and yoyo() methods of the Tween class are legacy and will be deprecated and removed.

One point of Timeline is to replace these features, so that Tween is cleaner, and orchestration is done cleanly via Timeline.

For example to repeat tweens, we can clone them and add multiple tweens to a timeline (do not make any code changes for cloning a Tween at the moment).

For example for delay, we can currently add a tween at an offset in a timeline.

For example for yoyo, we should be able to simply reverse a Tween and add that to a Timeline (do not make code changes for reversing a Tween at the moment).

For now, delete those methods in Timeline that were borrowed from Tween (do not modify Tween yet). These methods are unnecessary for Timeline and only give us extra complexity that we will solve with simpler Tweens and Timelines orchestration.

Finally, make sure the Timeline documentation clearly describe the features (except no string DSL, no methods borrowed from Tween).

trusktr (changes requested):
@copilot implement requested changes.

src/tests.ts
```diff
@@ -2904,29 +2910,74 @@ export const tests = {
...
 
 		tl.update(800)
 		test.equal(obj3.z, 100)
 
-		// '<' means start of last added, '>' means end
-		const tl2 = new TWEEN.Timeline()
-		const a = new TWEEN.Tween({x: 0}).to({x: 1}, 400)
-		const b = new TWEEN.Tween({x: 0}).to({x: 1}, 400)
-		tl2.add(a, 100)
-		tl2.add(b, '<')
```
trusktr: @copilot you removed this, but should add an equivalent example if you haven't already. 

This one can be `tl2.add(b, a)`.

Make sure all tests/examples you removed in the last commit have equivalent new tests/examples if needed.

trusktr (changes requested):

trusktr:
@copilot explain how time tracking while a Timeline is playing currently works.

dalisoft:
Hey @trusktr — pushed the follow-up for your direction. Tween `delay`/`repeat`/`repeatDelay`/`yoyo` are deprecated but still working; Timeline replaces them with `add(tween, {repeat})` / `{yoyo}`, backed by new `tween.clone()` and `tween.reverse()` (clone with mirrored progress, so an exact time-reverse with no snapshotting). Adding the same tween twice auto-clones, nested timelines repeat via `Timeline.clone()`. Guardrails: positive-integer repeat, 72h expansion cap. 1057 assertions green, guide updated, new `22_timeline_repeat.html` example. `undo()` stays for a later PR.

Two questions: is the clone-plus-mirror `reverse()` what you had in mind, and should yoyo on a nested timeline throw or wait for a future `Timeline.reverse()`?

@copilot-swe-agent — please review, focusing on the reverse/legacy-yoyo interplay, the `_addSingle` expansion, and the cap placement. Thanks!

trusktr:
@copilot can you able to update the description?

it says

> Breaking: Tween.yoyo() now throws (was deprecated-then-removed anyway); Group types widen Tween -> Tween|Timeline (runtime unchanged for Tween-only code). Version stays 25.0.0 (non-breaking; minor on release).

yoyo does not now throw, it is deprecated, otherwise that'd be breaking which is in conflict with releasing a non-breaking update.

src/Group.ts
```diff
@@ -84,12 +89,15 @@ export default class Group {
...
 		}
 	}
-	onComplete(callback: (object: Tween[]) => void) {
+	onComplete(callback: (object: Array<GroupChild>) => void) {
 		const group = this.getAll()
 		group.forEach(tween => {
-			const prevCallback = tween.getCompleteCallback()
-			tween.onComplete(() => {
-				prevCallback?.(tween)
+			// Timelines have onComplete but no getCompleteCallback; only wrap when available (plain Tweens).
```
trusktr: Lines 95-100: @copilot this is undesirable type casting. A Timeline is not a Tween. Fix it.

trusktr (changes requested):
@copilot

src/Group.ts
```diff
@@ -84,15 +89,26 @@ export default class Group {
...
+			if ('getCompleteCallback' in child) {
+				const prevCallback = child.getCompleteCallback()
+				child.onComplete(object => {
+					prevCallback?.(object)
+					notifyIfComplete()
+				})
+				return
+			}
+
+			child.onComplete(() => {
```
trusktr: This overrides any user oncomplete callback. Add getOnCompleteCallback to Timeline just like Tween's, and clean up this code here.

trusktr (changes requested):
@copilot please fix

src/Group.ts
```diff
@@ -98,16 +98,18 @@ export default class Group {
...
+					prevCallback?.(timeline)
 					notifyIfComplete()
 				})
 				return
 			}
 
-			child.onComplete(() => {
+			const prevCallback = child.getCompleteCallback()
+			child.onComplete(object => {
+				prevCallback?.(object)
```
trusktr, line 103: @copilot undo the conditional branching. If both Tween and Timeline have the exact same API for complete callbacks, this code is pointless. The code should be very similar to the original  before the PR with both tweens and timelines having the same oncomplete API

trusktr (changes requested):
@copilot fix

trusktr:
@copilot  don't make changes without making sure tests pass

trusktr:
> > @copilot  don't make changes without making sure tests pass
> 
> Confirmed on the latest change set: `npm run build` and `npm run test-unit` passed before I pushed `b01abcb`.

Test are failing. Fix.

trusktr:
@copilot move that last fix into a separate PR and delete the change from this PR.

trusktr:
@dalisoft copilot made a fix for timing in Tween.ts, and updated the unit test for end() which was failing (only sometimes?). 

I told copilot (here via GitHub) to move the changes to a separate PR but it failed miserably making new PRs. Needs cleanup. 

That was a test, and I learned I don't want to use copilot in GitHub so I'll use my own agent instead, it is way better because GitHub's copilot has too many limitations.

dalisoft:
@trusktr Yes, i am using OpenCode as Agent. Let me know if any update or fix needs. I will do update

trusktr:
@dalisoft I added e2e visual snapshot testing with playwright. You can see `npm run test-e2e` along with `playwright.config.ts` in the last commit.

Now fixing some issues, and adding more tests:

- Timeline restarts not resetting values. In the video, the 1st and 3rd boxes jump  unexpectedly.
  
  https://github.com/user-attachments/assets/92b10c7f-d6ae-4c01-b275-6f760eb3f020
- With the restart issue fixed, in the repeat example the second example (yoyo) gets out of sync with the 3rd example (yoyo+repeat) after a while.
  
  https://github.com/user-attachments/assets/c9254c9f-b08a-4119-9ab4-47b97b06fbfd

I'll push up new e2e tests for timeline in just a bit.

dalisoft:
@trusktr I look into issue, as tomorrow i will try to fix ASAP

trusktr:
@dalisoft Already fixed! That's what I meant by "now fixing". :) See last commit with updated tests.

However there are still more edge cases to create and test. When you get a chance you can make cases, for example multiple timelines inside timelines, yoyo and repeat timelines, etc.

Make sure you check output and don't allow redundant code. For example this range of changes removes redundant/useless code:

https://github.com/tweenjs/tween.js/compare/e224f04..e39e457

dalisoft:
@trusktr Today was busy, either tomorrow or after tomorrow i will rewrite some logic manually to reduce overhead and keep performance or improve

PR #711, opened 2026-08-16
Add new easing function generators: generateBack, generateElastic, generateSteps

0xTHAC0:
## Summary

Adds three new configurable easing generators to complement the existing `generatePow`, as requested in #507:

- **`generateBack(overshoot)`** — Back easing with custom overshoot (default `1.70158`, matching the built-in `Back`). Higher values produce more dramatic overshoots.
- **`generateElastic(amplitude, period)`** — Elastic easing with configurable amplitude (≥1) and period. Useful for springy animations where you need fine control over the bounce.
- **`generateSteps(steps)`** — Discrete step easing that quantizes the animation into N steps. In variant jumps at the start of each interval, Out at the end, InOut combines both halves. Useful for sprite sheet animations and staircase effects.

All generators return `EasingFunctionGroup` (In/Out/InOut) and satisfy the standard edge-value contracts:
- `f(0) = 0`
- `f(1) = 1`
- `InOut(0.5) = 0.5`

### Changes
- `src/Easing.ts` — added the three generator functions
- `src/tests.ts` — added 7 test cases covering edge values, discrete step behavior, midpoint constraints, and `generateBack` parity with default `Back`
- `examples/03_graphs.html` — added visual demos for the new generators

### Usage

```js
import { Easing } from '@tweenjs/tween.js'

// Back with stronger overshoot
const strongBack = Easing.generateBack(3)
tween.easing(strongBack.InOut)

// Elastic with more bounce
const springy = Easing.generateElastic(1.5, 0.4)
tween.easing(springy.Out)

// 4-step staircase
const steps = Easing.generateSteps(4)
tween.easing(steps.In)
```

Closes #507

PR #718, opened 2026-09-19
Easing function generators

trusktr:
continued from 

- #711


trusktr:
We'll bring 

- https://github.com/tweenjs/tween.js/pull/619

into here.

